PR #14, opened 2018-07-01
Line endings

selmf:
This PR converts YACReader's repository to use LF line endings internally. Applying it is perfectly safe and it should help us avoid cross-platform issues.

PR to master as we want to enable this ASAP.


selmf:
Ok, I just rebased this to resolve the conflict with the last merge. @luisangelsm please look into merging this as soon as possible (I don't want to constantly rebase/update this patch). You can use https://github.com/YACReader/yacreader/pull/14/files?w=1 for reviewing, this link will only show you the non-whitespace changes.

PR #26, opened 2018-08-05
qml line endings

selmf:
This PR fixes the line endings for files that were forgotten in the previous PR.


.gitattributes
```diff
@@ -59,6 +59,7 @@
 *.h text
 *.h++ text
 *.hh text
+*.qml text
```
luisangelsm: We also need `.pro`, `.pri` and `.qrc` files, and maybe the `.sh` files too.  Not sure if that's all...

selmf:
Ok, this should now cover almost everything. You can use this command to search for CRLF line endings:

$ git grep -Il $'\x0D'

This searches the git database, but excludes binary files.

PR #30, opened 2018-08-31
Fixed potential stack overflow.

jlesage:
In my case, I am compiling YACReader against `musl` libc, under Alpine Linux.

`musl` is known to have a stack size much smaller than `glibc`, thus explaining why the issue is not seen on most platforms.


selmf:
Looks OK to me, but it might take a little while until @luisangelsm merges it. @jlesage, could you change the target branch to develop?
I'd also be interested in some additional information on how you stumbled upon this problem so we can check if we have more potential stack overflows.

jlesage:
So I am building `YACReaderLibrary` for Alpine Linux.  During my tests, I tried the iOS application to test the import feature.  Every attempt to read or import a comic resulted in a crash of the application.  I just used GDB to get the backtrace to find out which function was crashing.  Then, looking at the source code, I quickly saw the huge buffer allocated on the stack.

PR #36, opened 2019-01-25
Feature/extended search

iain-benson:
Extend the search facility to allow Boolean logic, and a lot more fields to be searched.


selmf:
Hi @iain-benson,

thanks for the PR. As you might have noticed, development is a bit slow these days, so it could take a while before @luisangelsm can check the code.

There's also some stuff in the code that will likely require some discussion, such as switching to C++17 and adding an embedded library to the codebase.

luisangelsm:
First of all, thank you very much for the PR, it is really appreciated. This is something that has been in the TODO list for so long that it is great to finally see something done about it.

I didn't have time to fully review this yet, but I have the same concerns expresed by @selmf. @iain-benson could you post here an example of a search string? I would like to think about the posiblity of write a small class to take care of the lexical parsing, it shouldn't be hard and it will be easier to mantain in the long term.

Another thing that comes to my mind is that the fields used in the query parser should be defined in a single place only, so maybe it is time to do something in the database source code so these fields can be imported directly into the query parser, so further changes/updates in the database won't requiere updating the query parser fields manualy.

luisangelsm:
A couple of things need to be done before merging this:

1. Move the queries execution to a background thread. Now that the query is more complex, execution time can be high in large comic libraries, blocking the main thread leads to a bad user experience. Probably an execution queue is in order so we can cancel operations while the user types. This is something I wanted to do even before this PR.
2. Write a recursive descent parser and remove the `lexertl` dependency, the grammar is so small and simple that I think is hard to justify adding a dependency to the project.

I will take care of this at some point, if anyone is interested please ping me.

iain-benson:
Apologies for having somewhat abandoned this, I just haven't had the time to revisit it, other than to fix an issue I found with string streams after moving to a newer version of Xcode.

I understand why you want to get rid of `lexertl`, it probably is overkill, it was just quicker, and easier for me to find a parser than try to write my own.

Fixing the input lag on queries, either by threading, or possibly a simple hold-off whilst still typing, was something I had realised needed doing, but never quite got around to, I was also planning to implement query operators (as with the syntax at https://manual.calibre-ebook.com/gui.html#the-search-interface) rather than everything being `LIKE` or exact matches, but as with most things, life and work have a tendency to get in the way.

luisangelsm:
@iain-benson no worries :) Even if you can work any more on it, it is a good starting point. An user expresed interest in continue the work, but I didn't hear anything else from him, so I am not sure about the current status.

luisangelsm:
There have been some changes to the DB code and this PR was conflicting with the current implementation. I rebased it and fixed the conflicts, hopefully I didn't break anything.

I'll try to finish otherwise it won't survive as conflicts will get harder to fix :/ and it is a feature that has been requested a lot.

iain-benson:
Thank you for finishing this off. I knew there were issues with it stopping it being accepted, I just haven’t had the opportunity to look at it.

luisangelsm:
@iain-benson thank you for starting it :), it needs some more work though, requesting data from the DB should be done in a background thread. I am working on it now and hopefully I'll be able to merge it in a few a days.

Now that I have your attention, do you remember what `atWord` tokens are for?

iain-benson:
> @iain-benson thank you for starting it :), it needs some more work though, requesting data from the DB should be done in a background thread. I am working on it now and hopefully I'll be able to merge it in a few a days.
> 
> Now that I have your attention, do you remember what `atWord` tokens are for?

I don’t think it’s needed. I think it’s part of a feature used in Calibre’s python code, related to custom database fields, that I mistakenly put in despite not being used.

luisangelsm (approved):
It could use some more work but it seems to work good enough now.

YACReaderLibrary/db/comic_model.cpp
```diff
@@ -595,59 +596,25 @@ void ComicModel::setupReadingModelData(const QString &databasePath)
...
+    qDeleteAll(_data);
 
-        default:
-            QLOG_ERROR() << "not implemented";
-            break;
-        }
+    _data.clear();
 
-        selectQuery.exec();
+    QLOG_ERROR() << "-d2>" << data->size();
```
vedgy, line 609: Was this line left here by mistake? It spams error messages every time the search edit's text changes to a non-empty value.

luisangelsm: Yes, my bad.

vedgy: Should this line be removed or converted to `QLOG_TRACE` or `QLOG_DEBUG`?

luisangelsm: Removed, I used that log to catch some issues while I was working on this PR but it doesn't have a purpose anymore.

vedgy: Removed in #194.

PR #40, opened 2019-03-22
Fix gcc8 warnings

mytskine:
This branch fixes all the warnings of gcc/g++ when compiling on Linux.

There is at least one real bug-fix among these changes: in `YACReader/render.cpp` a line was obviously meant to be inside an `if` block, but was not because there was no {} block. The indentation was misleading. See 42abb93603a46f225233c2ea1ee84d760047905e.

Please don't feel blamed, but the indentation is awfully inconsistent and really error prone. Sometimes it's tabs, sometimes 4 spaces, and sometimes 2 spaces. I recommend cleaning up the entire codebase with an automatic tool like `clang-format`. Of course, all the MR and branches should be merged before this, since it would create conflicts everywhere.

```
% g++ --version
g++ (Debian 8.3.0-2) 8.3.0
```

### DIFF
--- a/YACReader/render.cpp
+++ b/YACReader/render.cpp
@@ -398,7 +398,7 @@ void PageRender::run()
 //-----------------------------------------------------------------------------
 
 Render::Render()
-:currentIndex(0),doublePage(false),doubleMangaPage(false),comic(0),loadedComic(false),imageRotation(0),numLeftPages(4),numRightPages(4)
+:comic(0),doublePage(false),doubleMangaPage(false),currentIndex(0),numLeftPages(4),numRightPages(4),loadedComic(false),imageRotation(0)
 {
 	int size = numLeftPages+numRightPages+1;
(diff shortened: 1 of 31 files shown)


selmf:
Oh wow. I don't think anyone feels blamed here, we are well aware of the shortcomings of the codebase. I'm partly to blame for adding and removing some of them myself ;)

I have two issues with the PR, though …

1. It should be done against the 'develop' branch (master is for stable releases and urgent bugfixes)
2. The certain bugfix you mention should be done in a separate PR

Just to be clear, I really appreciate this effort. I have tried something like this myself, but only succeeded in fixing some problems, but not eliminating all warnings.

mytskine:
I extracted the fore mentioned bug, and another one, into distinct PR. I pushed a new branch that replaces this one, so I'm closing this one.

I suggest you should edit the readme and add a "contributing" section that warns about the "develop" branch.

PR #41, opened 2019-04-01
bug: 'else' without braces but several lines indented

mytskine:
The indentation suggests that the line that update `previousIndex` was meant to be inside the `if`, but the lack of braces means it isn't. I don't think it was an indentation error.

This bug was discovered while fixing the warnings of gcc compilation. See #40.

### DIFF
--- a/YACReader/render.cpp
+++ b/YACReader/render.cpp
@@ -1035,6 +1035,7 @@ void Render::updateBuffer()
 		}
 	}
 	else //add pages to left pages and remove on the right
+	{
 		if(windowSize<0)
 		{
 			windowSize = -windowSize;
@@ -1060,6 +1061,7 @@ void Render::updateBuffer()
 			}
 		}
 		previousIndex = currentIndex;
+	}
 }
 
 void Render::fillBuffer()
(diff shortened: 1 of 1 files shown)


selmf:
Thanks for taking the time to separate this from the rest of the bugfixes. The function in question is, if I recall this correctly, responsible for filling the prerendered page buffer as you navigate through the comic.
The mistakingly indented line updates the previous index variable, which kind of makes sense for both navigation cases (left and right), so I am not sure if this is really a bug. I will need to have a closer look at our usage of those index variables before I can determine what to do here. 

@luisangelsm, what is your take on this?

luisangelsm:
@mytskine first of all thanks for these PRs, I love to see people being part of the project. @selmf is the only long term contributor to YACReader and he has been doing a terrific job, he has lots of experience in the project and knows the code well enough, so you can take his comments as if they were mine.

About the PR, I don't think the branches are needed, we need to update `previousIndex` no matter the navigation direction. Once the buffer is updated we don't need any other page renderers, so a second immediate call to `updateBuffer` should do nothing, in other words, the window size to populate the buffer should be 0.

Sorry for the bad indentation, the project suffered from a severe tab/spaces indentation inconsistency back in the day after switching from Visual Studio to QtCreator.

PR #43, opened 2019-04-01
Fix gcc8 warnings

mytskine:
This branch fixes the compilation warnings of gcc. This replaces #40 (the context is described there).

As requested, I've moved some bugs discovered here into separated branches : #41 and #42.
I've also changed my commit policy to make micro-commits that should be easier to rebase.


luisangelsm:
Awesome, thanks!

PR #44, opened 2019-04-02
Wrong file end of line encoding and utf8 fix

tim77:


selmf:
Thanks for the pull request . We already fixed the line endings in the develop branch by setting default gitattributes, so there is no need to manually correct this, it will be gone when we merge the branch. 

Utf8 encoding is a different story though. We likely still have a good amount of those errors from the time when @luisangelsm wrote the comments and changelog in spanish. If you want to help getting rid of these we certainly wouldn't say no ;)

selmf:
Thanks for the pull request . We already fixed the line endings in the develop branch by setting default gitattributes, so there is no need to manually correct this, it will be gone when we merge the branch. 

Utf8 encoding is a different story though. We likely still have a good amount of those errors from the time when @luisangelsm wrote the comments and changelog in spanish. If you want to help getting rid of these we certainly wouldn't say no ;)

tim77:
Ah, great. `YACReader` will be in official repos soon BTW (i hope). 😊

selmf:
Ah, great! Which distribution?

tim77:
Apologize, even forgot to mention this. 😄 In Fedora repos.

PR #57, opened 2019-05-24
Miscellaneous code fixes

vedgy:
See details in the commit messages.

### DIFF
--- a/YACReader/goto_flow.cpp
+++ b/YACReader/goto_flow.cpp
@@ -260,6 +260,8 @@ PageLoader::PageLoader(QMutex * m):
 
 PageLoader::~PageLoader()
 {
+	//TODO this destructor never runs. If it is ever called, it will hang, because
+	//the implementation is broken due to the absolutely endless loop in run().
 	mutex->lock();
 	condition.wakeOne();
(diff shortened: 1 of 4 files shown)


YACReaderLibrary/library_window.cpp
```diff
@@ -1798,8 +1798,6 @@ void LibraryWindow::checkEmptyFolder()
 void LibraryWindow::openComic(const ComicDB &comic)
 {
     if(!importedCovers) {
-        QList<ComicDB> siblings = comicsModel->getAllComics();
```
luisangelsm: Good catch, this is now done in YACReader, it request the siblings to the local server in YACReaderLibrary.

YACReader/goto_flow.cpp
```diff
@@ -260,6 +260,8 @@ PageLoader::PageLoader(QMutex * m):
 
 PageLoader::~PageLoader()
 {
+	//TODO this destructor never runs. If it is ever called, it will hang, because
```
luisangelsm: I am fine with this TODO, goto_flow (non gl) is or should be marginally used.

Issue #56, opened 2019-05-24
YACReader often uses an entire CPU core from launch till exit if hardware acceleration is disabled


vedgy:
Steps to reproduce:

0. Disable hardware acceleration in YACReader options.
1. Open a comic book (cbr or cbz). The bug happens more often for large comics (>=100 pages). Also seems to happen more often when opening by double-clicking in YACReaderLibrary. A comic book never opened in YACReader before is preferable.
2. If the YACReader CPU usage goes down to 0-1% after a few seconds, exit YACReader and open the same comic again. This time the CPU usage will likely stay high, e.g. 25% on a 4-CPU-core system.

I tried to debug the issue in Qt Creator by interrupting YACReader built in Debug mode while it was using the entire CPU core. But the call stack didn't indicate a problem in any thread. Tried QML debugging in Qt Creator, but it didn't work on my system for some reason.

System and versions: Manjaro Xfce x64 with latest updates; Qt 5.12.3; https://aur.archlinux.org/packages/yacreader with `CONFIG+=no_pdf` instead of `CONFIG+=pdfium`.

EDIT: the same issue happens when compiling the *develop* branch in Qt Creator using the default *poppler* PDF backend.

selmf:
When you disable the hardware acceleration you enable software rendering for the cover flow. This will naturally use more CPU than hardware rendering, nothing strange about that. This does not affect QML views.

When you open a comic for the first time, YACReader will extract or render its first page. When you open a comic for the second time, it will try to open the last page you viewed, which can need more CPU, depending on the format used.

YACReader will also extract or pre-render the whole comic you opened, starting from the last page you viewed. For some formats, this can generate extra CPU load as some pages are extracted twice (one pass to reach the last page you viewed, second pass to prepare the pages before that page).

And last but not least, YACReader will also create page thumbnails for the cover flow. That also uses CPU and scales with the length of the comic loaded.

The reason for these strategies is to display a page as soon as possible and to buffer the comic for smooth reading. It works very well with files in the 10 to 100 mb range, but with gigabyte size comics it gets a little memory and cpu hungry.

Btw, if you want to see debug output, the develop version of YACReaderLibrary has a command option to set a loglevel on launch. Older versions suffer from a bug where some of our third party code installed a global debug handler which ate all the output without displaying it.

vedgy:
[...]

PR #58, opened 2019-05-25
Fix a data race in PictureFlow::triggerRender()

vedgy:
`PictureFlow::triggerRender()` is called from at least two different threads. Thus modifying `triggerTimer` in this function was a data race.

Not sharing the timer object should not degrade performance because the documentation for `QWidget::update()` [states](https://doc.qt.io/qt-5/qwidget.html#update):

> Calling `update()` several times normally results in just one `paintEvent()` call.

Without this fix YACReader on GNU/Linux often uses an entire CPU core from launch till exit if hardware acceleration is disabled (#56).

This commit also eliminates the following errors from YACReader's standard output:

>     QObject::killTimer: Timers cannot be stopped from another thread
>     QObject::startTimer: Timers cannot be started from another thread

The code requires Qt 5.4 or later, which should be OK as INSTALL.md [lists](https://github.com/YACReader/yacreader/blob/master/INSTALL.md#build-dependencies) "Qt >= 5.6" as a build dependency.

### DIFF
--- a/common/pictureflow.cpp
+++ b/common/pictureflow.cpp
@@ -999,7 +999,6 @@ public:
   PictureFlowState* state;
   PictureFlowAnimator* animator;
   PictureFlowAbstractRenderer* renderer;
-  QTimer triggerTimer;
 };
 
 
@@ -1034,8 +1033,6 @@ PictureFlow::PictureFlow(QWidget* parent,FlowType flowType): QWidget(parent)
   d->animator = new PictureFlowAnimator;
   d->animator->state = d->state;
   QObject::connect(&d->animator->animateTimer, SIGNAL(timeout()), this, SLOT(updateAnimation()));
-
-  QObject::connect(&d->triggerTimer, SIGNAL(timeout()), this, SLOT(render()));
 
 #ifdef PICTUREFLOW_QT4
   setAttribute(Qt::WA_StaticContents, true);
@@ -1194,13 +1191,7 @@ void PictureFlow::render()
 
 void PictureFlow::triggerRender()
 {
-#ifdef PICTUREFLOW_QT4
-  d->triggerTimer.setSingleShot(true);
-  d->triggerTimer.start(0);
-#endif
-#if defined(PICTUREFLOW_QT3) || defined(PICTUREFLOW_QT2)
-  d->triggerTimer.start(0, true);
-#endif
+  QTimer::singleShot(0, this, &PictureFlow::render);
 }
 
 void PictureFlow::showPrevious()


vedgy:
After some more digging I no longer believe that this was a data race in `PictureFlow`. Either there still is a data race in *goto_flow.cpp* or the issue was not a race, but calling `QTimer::start()` from another thread, which is prohibited by the `QTimer` API. I'm not sure if `GoToFlow` and `SlideInitializer` access their shared `PictureFlow` object in a thread-safe manner.

By the way, there is a memory leak in *goto_flow.cpp*: a new `SlideInitializer` is constructed each time a comic is opened in YACReader and never destroyed.

`PageLoader` is also never destroyed, but at least it is created only once in the YACReader process lifetime. Still such leaks are detected by valgrind. If there are many such intentionally never destroyed objects, finding real memory leaks becomes more difficult.

vedgy:
There **is** a data race in *goto_flow.cpp*. I added `QThread::sleep(100);` before the loop in `SlideInitializer::run()`. `GoToFlow` then called non-const functions on `flow` without waiting for `SlideInitializer` to finish. The Page Flow was empty in the UI.

I think the code in `SlideInitializer::run()` is not slow enough to justify creating a dedicated thread. I tested several 1 GB comic books with 550 pages each. `SlideInitializer::run()` run for up to 8 milliseconds the first time a comic was opened; up to 25 milliseconds the second time and up to 6 milliseconds the third time the same comic was opened. For smaller 30-page comics the time is usually 0 milliseconds.

How about moving the code from `SlideInitializer::run()` into `GoToFlow::setNumSlides()` directly and deleting the `SlideInitializer` class?

selmf:
If I'm not mistaken we're not using the image loader and slide initializer classes with the thread problems in our OpenGL flow. @luisangelsm , is there any chance that we could reuse or port the image loader code we're using for the hardware acceleration codepath for the pictureflow? I'd rather not have two solutions in place, especially if one of them is already suffering from bitrot and seldomly used.

vedgy:
The tiny `SlideInitializer` class is unique (no corresponding OpenGL class) and used only when hardware acceleration is disabled. `GoToFlow` doesn't deal with threads on its own, and I doubt there are any thread safety bugs to be fixed inside this relatively large class.

There are 4 *very* similar worker classes: `PageLoader` in *goto_flow.\**, `ImageLoaderGL` and `ImageLoaderByteArrayGL` in *yacreader_flow_gl.\**, `ImageLoader` in *comic_flow.\**. All 4 classes are not completely thread safe for at least one reason: their non-atomic data member `working` is modified in their own thread and accessed without locking from an external thread in the public `busy()` member function. I'm pretty sure this is not the only bug in all them. I think it's a good idea to unify these 4 classes into one class template. And then fix the implementation of this class template.

My brief investigation makes me think that unifying all the OpenGL and non-OpenGL *flow* code will be difficult, time-consuming and possibly not worth doing at all. @luisangelsm's opinion on this is much more authoritative of course. If no one is ready to do such a large-scale refactoring in the nearest future, I can take these much simpler steps for now:
1. Get rid of the `SlideInitializer` class.
2. Unify the 4 similar loader classes and fix the new class template.

selmf:
Unifying the image loaders and possibly porting them to something a little more maintainable sounds like a good idea to me. There are probably even more places where such a class could be used, I.e  in the library and it our render code for the viewer (not to be confused with the flow code!).  If you want to tackle this you're more than welcome, but you'll need a good deal of persistence. @luisangelsm is very picky when reviewing code that touches threads (for a good reason) and his work and life schedule does not always give him enough time to do a proper review. You may also want to ask him for a slack invite so we can discuss technical details.

luisangelsm:
I need to read all the code again before making a decision.  

The loading system was taken and modified from the original picture flow code and used later in all the "flows".

The access to `working` is read only, the only risk of using it is not being accurate when preloading is triggered I mean, there is no possible data corruption here, I think (iirc) that it is queried from the main thread, so we don't really want to block it.

I am not agains any refactorings as far as we don't end worse than the current situation. When a problem is generalized I get tons of activity in the mail and the forum and so far this is not the case. Probably it would be better to have CI working before doing huge refactors, so we can easily offer nightly builds and get a good number of users for testing.

Did you actually find where/when this 100% single core usage happens?

vedgy:
> The access to `working` is read only, the only risk of using it is not being accurate when preloading is triggered I mean, there is no possible data corruption here, I think (iirc) that it is queried from the main thread, so we don't really want to block it.

The data race is undefined behavior and I think it should be fixed. It is quite enough to worry about thread safety, no need to also consider whether or not this undefined behavior is "safe" in each particular situation. From a non-top answer to https://stackoverflow.com/questions/9200951/is-it-ok-to-read-a-shared-boolean-flag-without-locking-it-when-another-thread-ma:

> If you're using C++11, benign races are undefined behavior. Even when undefined behavior doesn't make sense for the underlying data type. The problem is that compilers can assume that programs have no undefined behavior, and make optimizations based on that...

-------------
> Did you actually find where/when this 100% single core usage happens?

It seems that the one-core CPU usage is caused by some Qt event endless loop. I think the event loop is stuck because of a data race in `QTimer` code caused by YACReader's violation of its API (e.g. the `SlideInitializer`'s thread may try to restart the timer while it emits a signal or is otherwise modified in the main thread's event loop).

When I pause debugging while the bug is being reproduced, I see different call stacks in the main thread with only the bottom-most function inside the YACReader code. For example (note `QTimerInfoList::activateTimers()`):
```
1  QObject::event(QEvent *)
2  QApplicationPrivate::notify_helper(QObject *, QEvent *)
3  QApplication::notify(QObject *, QEvent *)
4  QCoreApplication::notifyInternal2(QObject *, QEvent *)
5  QTimerInfoList::activateTimers()
6  ??
7  g_main_context_dispatch
8  ??
9  g_main_context_iteration
10 QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>)
11 QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>)
12 QCoreApplication::exec()
13 main                                  main.cpp 136
```

When I fix the bug either with ea299820fc01377eff82459a8b59c881c146d5e2 or by eliminating the `SlideInitializer` class, the call stack in the main thread always ends with `poll` after the initial loading is done and the CPU usage goes down to zero:
```
1 poll
2 ??
3 g_main_context_iteration
4 QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>)
5 QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>)
6 QCoreApplication::exec()
7 main                                  main.cpp 136

```

Once @luisangelsm makes a decision, I'd like to start with the first step - *Get rid of the SlideInitializer class*. This is an easy to implement, review and merge fix. And it is practically independent from the more involved refactoring and fixing of the loader classes.

selmf:
That would roughly correspond with what I could see when testing. Lots of time spend in the event loop, polling shared library, and sometimes a qwaitcondition error on exit. I also tried to narrow it down by naming threads and adding debug messages, but this didn't give me any usefull information.

luisangelsm:
Ok, I think we can merge this and left `SlideInitializer` there (I would like to find and read some commits about that class before deleting it), I would include signal/slot connection for deleting the SlideInitializer once it is done (finished -> deleteLater) and change `working` type to `std::atomic<bool>`.

We can think about a deeper refactoring later, once the format is fixed using `.clang-format` I want to go throw the code and fix a good amount of warnings and remove/fix/update some comments and TODOs. That should left us a cleaner code base to work on.

@vedgy if you want to discuss anything I can invite you to YACReader's slack, so we can talk and resolve questions and doubts about the code without the need of creating issues or PRs

vedgy:
1. **`SlideInitializer`.** I am not sure if this pull request is worth merging. The issues with `SlideInitializer` go beyond the memory leak and the timer issues. As I described in the just created #59, thread synchronization between `SlideInitializer` and `GoToFlow` is missing entirely.

2. **`PageLoader`.** Making `working` atomic is a good idea, but I'm pretty sure it won't make `PageLoader` entirely thread-safe. For example, see my third commit - 12aabf79603a58d13678f0576d15ad2a57fa46e3 - in #57. The fixes in this commit protect the shared `restart` variable and follow standard `condition_variable` documentation. Of course, the code actually uses `QWaitCondition`, which doesn't suggest `while` over `if`. But I found [this question](https://stackoverflow.com/questions/55100268/can-qwaitcondition-spuriously-wake-up) with a comment suggesting that `while` is always safer. However I highly doubt that this superficial commit will fix all the remaining issues with `PageLoader`. Still, if you are not ready for refactoring of this and the other loader classes, it can wait for a while since there are no associated user-reported bugs.

3. **Slack.** I don't have a Slack account. And I don't like the idea of private code discussions. Such discussions are more likely to be difficult to access or lost eventually. If there are code rules or guidelines, they should be in the open rather than hidden in private Slack discussions - for the benefit of future contributors. I think that refactoring plans and prioritization should also be public for the same reason. In addition, there are privacy concerns: https://en.wikipedia.org/wiki/Slack_(software)#Criticism. What's the problem with pull requests anyway? :)

vedgy:
Let me elaborate why I don't want to merge this `PictureFlow` pull request into the `develop` branch.

The change makes sense *only* if `PictureFlow` is used from more than one thread. However it is clearly not meant to be used like this as there is absolutely no multi-threading or thread synchronization code in the implementation of `PictureFlow`. If there is a need to use the class from multiple threads in the future, either a new class should be made to inherit `PictureFlow`, or specifically documented member functions, safe to call from non-main threads should be added into `PictureFlow`.

I consider this pull request a workaround for the high CPU usage issue. And it does have downsides: there is some overhead of constructing multiple `QTimer` objects and calling `PictureFlow::render()` more times. So this workaround can prove disadvantageous and worth reverting in the long term.

On the other hand, getting rid of `SlideInitializer` eliminates several problems at once. If there actually is a minor performance benefit of this class, a safe solution could be implemented in the future when there is time for such optimization.

> (I would like to find and read some commits about that class before deleting it)

Please check out those commits. I don't have access to them because the commit that added `SlideInitializer` is missing from this public git repository (or this addition never was in a separate commit).

luisangelsm:
Ok, I am closing it then.

I am fine with eliminating `SlideInitializer`, changes on the non gl version of the flow won't have much impact, and a few ms won't be noticeable (maybe a hiccup with ultra large comics and super slow machines).

selmf:
Regarding the Slack invite... nothing is wrong with pull request, but a direct form of communication is often more effective. We're also not interested in making a new evet

vedgy:
I see. I may join Slack if I become more involved in YACReader development in the future. For now I have implemented a few fixes and small features, which I am currently testing. I'll rebase my commits and create pull requests once #60 is merged to avoid conflicts. After that there will be this refactoring and fixing of the loader classes, which I don't expect to be a huge undertaking (several days of interesting work I think). I don't feel like these contributions warrant joining Slack. Pull requests should be sufficiently effective. In case of the loader classes I'm going to start with WIP to get intermediate code feedback and to ask questions.

It's unfortunate that the conversion from Mercurial didn't preserve commit history.

luisangelsm:
Old repo is available here: https://bitbucket.org/luisangelsm/yacreader/src/develop/

selmf:
Thanks for the contributions, it is really appreciated. For the loader classes, there is a lot of interesting stuff available. One thing I had in mind for this was using QThreadPool to avoid spawning and despawning threads, but there was an issue that made me abandon it (I think what I had in mind would at least need Qt 5.9).

There is also one other thing I wanted to ask you @vedgy . I have a stale branch of YACReader where I rewrote the OpenGl code to use a modern render pipeline. Development has advanced to a point where this branch is no longer easily merge- or rebaseable, but I'm curious if [this](https://github.com/YACReader/yacreader/tree/opengl_refactoring) would be an improvement over your current problem with the OpenGL flow. My guess is "probably not" as I'm still creating textures and vertexes in a rather inefficient way, but it would be nice to have some feedback on this as I am slowly cherrypicking features from that branch over to our current develop branch.

vedgy:
I converted https://bitbucket.org/luisangelsm/yacreader/src/develop/ into a git repository, but `SlideInitializer` is in the very first commit - *commit inicial en mercurial*.

@selmf, I don't know much about OpenGL and am not interested in it. I haven't noticed any lags with software rendering. It's just images, not video. So I don't want to spend much time on merging or cherry-picking from your branch. And I don't want to test the branch so many commits behind master as it is (without merging).

PR #60, opened 2019-05-30
Add .clang-format and format all code

luisangelsm:


selmf:
Two nits:

- we don't use qdoc and translating comments, it probably would be better to remove this rule to avoid surprises
- ForEachMacros - using Q_FOREACH is deprecated/discouraged by Qt, c++11 for range loops are prefered. 

For the actual layout or style generated by this I still need to play with this a little to fully understand it. It also would be nice to have some input on this from @mytskine :)

luisangelsm:
Updated

PR #91, opened 2019-10-02
Fix the hyperlink to project releases in README

prantlf:
When clicking on the hyperlink at https://github.com/YACReader/yacreader, it leads to a 404.


luisangelsm:
Thanks!

PR #99, opened 2019-11-17
Unify image loader classes

vedgy:
This is the first pull request in a series of two. It introduces the common worker class and uses it instead of one of the 4 image loader classes. See commit messages for details.

Note: I don't like some of the formatting changes clang-format does. There aren't many templates, lambdas or default member initializers in YACReader code yet. Consider the results and see if you like them. If not, you could change some of the formatting rules. I'd be happy to reformat changes in this pull request afterwards :). These are clang-format changes to the initial version of this pull request: [clang-format-changes.patch.zip](https://github.com/YACReader/yacreader/files/3855411/clang-format-changes.patch.zip).

### DIFF
--- a/YACReaderLibrary/YACReaderLibrary.pro
+++ b/YACReaderLibrary/YACReaderLibrary.pro
@@ -107,6 +107,8 @@ HEADERS += comic_flow.h \
   ../common/comic.h \
   ../common/bookmarks.h \
   ../common/pictureflow.h \
+  ../common/release_acquire_atomic.h \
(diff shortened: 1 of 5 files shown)


vedgy:
There seems to be a problem with CI Windows builds:

> 'qmake' is not recognized as an internal or external command, operable program or batch file.

selmf:
Yes, there is an issue with the Qt install for the Windows builds. I'm looking into it.

selmf:
Build issue is fixed on develop, you might want to rebase :)

PR #100, opened 2019-11-19
Reader: fix restoring Info visibility

vedgy:
See commit messages for details.

There is one more place where a temporary `QTimer` can be refactored into `QTimer::singleShot`: [`LibraryWindow::toNormal()`](https://github.com/YACReader/yacreader/blob/36c1cdd2dac57225aa4b9743120c2c71f7f6fab5/YACReaderLibrary/library_window.cpp#L2057). I am not touching that code for two reasons:
1. I can not test it because I don't have access to Mac.
2. I doubt that it currently works as intended because [`YACReaderMacOSXToolbar::show()`](https://github.com/YACReader/yacreader/blob/36c1cdd2dac57225aa4b9743120c2c71f7f6fab5/custom_widgets/yacreader_macosx_toolbar.h#L59) is a simple function. As far as I know, only slots can be connected to with the old signal and slot syntax.

### DIFF
--- a/YACReader/main_window_viewer.cpp
+++ b/YACReader/main_window_viewer.cpp
@@ -1217,11 +1217,7 @@ void MainWindowViewer::checkNewVersion()
         connect(versionChecker, SIGNAL(newVersionDetected()),
                 this, SLOT(newVersion()));
 
-        auto tT = new QTimer;
-        tT->setSingleShot(true);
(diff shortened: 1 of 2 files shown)


luisangelsm:
Sorry it is taking me forever to merge this. Nice catch btw, I need to take a look to the macOS side and fix things there.

PR #103, opened 2019-11-26
Unbreak build on FreeBSD

jbeich:
`AF_INET` is in `<sys/socket.h>` [according to](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_socket.h.html) POSIX.


selmf:
Looks good to me.

PR #129, opened 2020-04-09
Update azure pipelines to work on forked repos

luisangelsm:
Other changes:

- Use macos catalina
- Fix Windows builds by including QtScript in Qt installation


selmf:
Looks good, but Windows pipelines still need QtScript to work

PR #157, opened 2020-09-02
Reader: Add option to stop enlarging images in fit to width and height

### LINKED ISSUE #110: [Feature request] Don't enlarge small images
At now, when you have turned on the option to fit image to height / width, small images becomes larger than their own sizes. We need an option to stop this behavior.

Don't enlarge small images more then their original size.

Small images are images that are smaller than the screen size.

selmf:
* Add a toggable setting for enlarging images smaller than screen or window in fit to width and fit to height mode.
* fixes #110


luisangelsm (approved):
Excelent

PR #165, opened 2020-09-23
Updated Dutch translation

Vistaus:
:-)


selmf:
Thanks for the update, it will be included in the next release :)

Issue #162, opened 2020-09-11
Please add a build for pre-10.13 macOS


jeremydouglass:
The proposal is for two YACReader builds, one bleeding edge and one legacy / LTS.

Contexts for this are the recent repo release minimum OS requirement for 9.7 increasing to macOS 10.13, as well as upcoming changes in the build process and user targets, including end of life for Qt5 and Apple's move to ARM based machines.

See previous discussion in https://github.com/YACReader/yacreader/issues/161#issuecomment-690941672

PR #166, opened 2020-09-24
Support for third party qt image plugins

smahot:
Hi,
There is a qt plugin for AVIF image format here : https://github.com/novomesk/qt-avif-image-plugin/
We just need to drop **qavif.dll** there : imageformats/qavif.dll
This plugin is build with Microsoft Visual Studio 2019 and Qt **5.14.2** so the qt version must be updated on the azure pipeline.
I updated it from **5.12.6** to **5.15.1** and now it works well with avif images !
May be, you can add this plugin as a dependencies to be bundled with the installer ?
Thanks !
Regards,
Steve

### DIFF
--- a/common/comic.cpp
+++ b/common/comic.cpp
@@ -23,6 +23,26 @@ enum YACReaderPageSortingMode {
 
 void comic_pages_sort(QList<QString> &pageNames, YACReaderPageSortingMode sortingMode);
 
+QStringList Comic::getSupportedImageFormats()
+{
+    QList<QByteArray> supportedImageFormats = QImageReader::supportedImageFormats();
+    QStringList supportedImageFormatStrings;
+    for (QByteArray item : supportedImageFormats) {
+        supportedImageFormatStrings.append(QString::fromLocal8Bit("*." + item));
+    }
+    return supportedImageFormatStrings;
+}
+
+QStringList Comic::getSupportedImageLiteralFormats()
+{
+    QList<QByteArray> supportedImageFormats = QImageReader::supportedImageFormats();
+    QStringList supportedImageFormatStrings;
+    for (QByteArray item : supportedImageFormats) {
+        supportedImageFormatStrings.append(QString::fromLocal8Bit(item));
+    }
+    return supportedImageFormatStrings;
+}
+
 const QStringList Comic::imageExtensions = QStringList() << "*.jpg"
                                                          << "*.jpeg"
                                                          << "*.png"
--- a/common/comic.h
+++ b/common/comic.h
@@ -76,8 +76,8 @@ public:
     //check if the comic has failed loading
     bool hasBeenAnErrorOpening();
 
-    inline static QStringList getSupportedImageFormats() { return imageExtensions; }
-    inline static QStringList getSupportedImageLiteralFormats() { return literalImageExtensions; }
+    static QStringList getSupportedImageFormats();
+    static QStringList getSupportedImageLiteralFormats();
 
     static bool fileIsComic(const QString &path);
     static QList<QString> findValidComicFiles(const QList<QUrl> &list);


selmf:
I don't see an issue with updating to a newer compiler and runtime. In fact this should be something we should automate as we regularly forget to do this for new releases.

Adding a file format that is not supported by standard Qt to the list of supported image formats is a no-go. The proper way to do this is to replace the fixed file format lists with QImageReaders's supportedImageFormats() function. See the Qt documentation for further details: https://doc.qt.io/qt-5/qimagereader.html

smahot:
Qt support by default these format : https://doc.qt.io/qt-5/qimage.html#reading-and-writing-image-files

Format | Description | Qt's support
-- | -- | --
BMP | Windows Bitmap | Read/write
GIF | Graphic Interchange Format (optional) | Read
JPG | Joint Photographic Experts Group | Read/write
JPEG | Joint Photographic Experts Group | Read/write
PNG | Portable Network Graphics | Read/write
PBM | Portable Bitmap | Read
PGM | Portable Graymap | Read
PPM | Portable Pixmap | Read/write
XBM | X11 Bitmap | Read/write
XPM | X11 Pixmap | Read/write

But it accepts third party codecs via plugins that are loaded dynamically inside /imageformats : https://doc.qt.io/qt-5/qtimageformats-index.html

Format | Description | Support | 3rd Party Codec
-- | -- | -- | --
ICNS | Apple Icon Image | Read/write | No
JP2 | Joint Photographic Experts Group 2000 | Read/write | Yes (Not bundled)
MNG | Multiple-image Network Graphics | Read | Yes (Not bundled)
TGA | Truevision Graphics Adapter | Read | No
TIFF | Tagged Image File Format | Read/write | Yes (bundled)
WBMP | Wireless Bitmap | Read/write | No
WEBP | WebP | Read/write | Yes (bundled)

Webp is bundled but not avif because the codec is new.
In the case you want to wait that the plugin is officialy bundled with qt, we still can let people to add it manually ?
By this way, people will just need to drop the dll inside /imageformats to get it working, isn't it ?

With the actual version 9.7.1, even if qavif.dll is there and is compiled with qt 5.12.6, the reader don't show the image as if it wasn't here.
But if i change the extension .avif to .jpg, it works.

selmf:
The reason I don't want to add .avif to the list of supported images is not that I am against supporting it but that it is generally a bad idea to use a static list in this case. As you have observed, you can add AVIF support via the plugin but the list effectively prevents you from using it. Yes, that can be fixed by adding ".avif" to the list or masquerading the images as something else. But that will only enable the AVIF plugin and might lead to undefined behaviour where it is not available.

The better approach is to get rid of the static lists and **ask Qt directly** what image formats are supported, including those installed via plugins. In other words, replacing the static image list with a dynamic list obtained via QImageReader will enable not only AVIF support but image support via plugins in general, without adding undefined behavior. It is thus a much cleaner and much more longterm solution.

smahot:
I see what you mean and you are right. I will check if i can load the list dynamically.

smahot:
Hi,
I reverted back the static list, it's not used anymore. I will let you clean the code.
Anywhere the static list were used, it call the function `QImageReader::supportedImageFormats()`, and format the `QList<QByteArray>` into the right format (`QStringList` or `QList<QString>`), and eventually add "*." before.
By default, `QImageReader::supportedImageFormats()` return 20 formats :
`bmp, cur, gif, icns, ico, jpeg, jpg, pbm, pgm, png, ppm, svg, svgz, tga, tif, tiff, wbmp, webp, xbm, xpm`
If the avif plugin is there, YACReader is able to load avif image. If the plugin is not here, the reader don't try to load the image, like before.
I wasn't able to see `avif` added into the list `QImageReader::supportedImageFormats()` in debug mode as the plugin is only loaded in release mode.
But in release mode, it works !

selmf:
Great to hear you're making progress! I'd suggest to leave the list creation at a central place, i.e. Comic::Comic::getSupportedImageFormats() and Comic::getSupportedImageLiteralFormats(), so we don't have multiple copies of the same code and can delete the old static lists. Qt's 'foreach' is deprecated, so C++11 ranged for loops might be a better idea.

Finally, for merging it would be good to have the plugin support as a separate PR without the changes to the CI.

smahot:
The list is now returned by `Comic::getSupportedImageFormats()` and `Comic::getSupportedImageLiteralFormats()`.
`Comic::imageExtensions` and `Comic::literalImageExtensions` are not used anymore.
And i also used a `for` loop instead of a `foreach` loop.
I'm not sure how to remove the CI update commit from the PR. Is it important ? I mean it's also needed for the support of the latest plugins.
If it is, please suggest me a clean way to do it :)

selmf:
Looks good! Thanks for going the extra mile. Keeping the changes in the CI apart from this PR is mostly best practices, as in "try to add one fix/feature per PR. We don't always stick to it, but we try.
To clean the PR, the best way would be to revert the git commit in question and then squash all commits on the PR into a single commit. That way I can cleanly merge your changes as a single commit on the DEV branch and we'll have it in the next point release.
The CI changes are something I would like to discuss in-depth with @luisangelsm first. Not because there is an issue with your proposed changes but because we need to consider a better way of handling transition to newer Qt releases and azure pipelines build images. This also touches issues like #168 and #162 and we also might consider building the AVIF plugin directly in our CI so it always matches whatever Qt version we are currently using.

smahot:
I squashed them and removed the CI update, it's cleaner now.
It's a great idea to build the avif plugin into CI ! It will be easier for the users.
I'm looking forward the next update :)

selmf:
Merged. Thanks you :+1:

smahot:
Thanks you too for your help and this software 👍

PR #174, opened 2020-10-25
Increase the maximum zoom level

### LINKED ISSUE #175: Increase the maximum zoom level
I download mangas for off-line reading. I save them as a directory of images, packed to a ZIP archive. Some mangas have a very small default size of the picture. Zooming to 500% of the original size is not enough to be able to read them. For most of them helps the magnifying glass, but it is inconvenient. I found, that the convenient zoom would have to be 2000%, at least.

**How about increasing the maximum zoom value to 3000%?**

It would make the zoom slider more difficult to use and that is why I stretched to be twice the original width and moved it a little to the left.

Thank you!

Examples:

Not zoomed:

not-zoomed

Zoomed to 500%:

zoomed-500

Zoomed to 2000%:

zoomed-2000

prantlf:
Increase the max zoom to 3000%, stretch the zoom slider to 400 and move it to the left by 150.

Attempts to fix #175.

### DIFF
--- a/YACReader/main_window_viewer.cpp
+++ b/YACReader/main_window_viewer.cpp
@@ -1419,7 +1419,7 @@ void MainWindowViewer::toggleFitToWidthSlider()
     if (zoomSliderAction->isVisible()) {
         zoomSliderAction->hide();
     } else {
-        zoomSliderAction->move(250, y);
+        zoomSliderAction->move(100, y);
         zoomSliderAction->show();
     }
 }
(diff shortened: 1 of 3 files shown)


selmf:
You are using the Zoom feature wrong. Zoom is meant to be applied as a modifier to the various image stretch modes YACReader supports. In your example you are either using "fit to height" or "fit to screen" while you should be using "fit to width" or "native resolution".
Fitmodes are accessible by clicking the the four pictograms left to the zoom button in the toolbar or by using the right-click context menu.

prantlf:
Yes, I always use fit to page and correct the size by zoom as needed. Thank you for letting me know about the proper usage!

Issue #112, opened 2020-02-09
Move up, Move down, Move left, and Move right are not configurable.


crocket:
In keyboard shortcuts > Reading, you can find Move up, Move down, Move left, and Move up.

Whether or not they are configured, they are always mapped arrow keys.
I cannot set them to other keys.

gavieira:
I've just tested this in the 9.6.2 version from the [AUR](https://aur.archlinux.org/packages/yacreader/) on Manjaro Linux 19.0.2 and had the same issue.

PR #176, opened 2020-11-06
Fixes move action shortcuts not working

smalewski:
Shortcuts for moving inside a page didn't work. Now they do.

### DIFF
--- a/YACReader/viewer.cpp
+++ b/YACReader/viewer.cpp
@@ -652,20 +652,20 @@ void Viewer::keyPressEvent(QKeyEvent *event)
         QAbstractScrollArea::keyPressEvent(event);
 }
 
-void Viewer::moveAction(QKeySequence key)
+void Viewer::moveAction(QKeySequence *key)
 {
     int _key = 0;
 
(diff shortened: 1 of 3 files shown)


smalewski:
Fixes this [issue](https://github.com/YACReader/yacreader/issues/112)

YACReader/viewer.cpp
```diff
@@ -634,20 +634,20 @@ void Viewer::keyPressEvent(QKeyEvent *event)
         QAbstractScrollArea::keyPressEvent(event);
 }
 
-void Viewer::moveAction(QKeySequence key)
+void Viewer::moveAction(QKeySequence *key)
```
luisangelsm, line 637: Did you try declaring this method as `void Viewer::moveAction(const QKeySequence &key)`? then you can invoke it as `moveAction(key)` and there is no need for indirection in the method's code. I think we can merge the PR after changing it.

luisangelsm (changes requested):

luisangelsm (approved):

luisangelsm:
@smalewski thank you very much for the PR, I am merging it now.

PR #187, opened 2021-01-25
Fix launch sequence to avoid http port collisions

luisangelsm:


selmf:
Looks good to me, but you might as well use the occasion to clean up the Spanish comments in the modified code.

selmf (approved):

selmf:
Gracias :)
I think it makes little sense to hunt after every legacy comment (or problem), but it is a good approach to fix these as we touch the code.

PR #189, opened 2021-01-30
Fix keyboard shortcuts with alternative layouts; always limit Magnifying glass height; clean up related code

vedgy:
In YACReader some shortcuts cannot be triggered with a non-Latin keyboard layout. For example: *Toggle fullscreen mode*, *Hide/show toolbar* and *Toggle between fit to width and fit to height* cannot be triggered when Ukrainian keyboard layout is active. The issue is that some actions are not connected to the corresponding slots and instead rely on custom `keyPressEvent()` handlers, which are not sophisticated enough to support different keyboard layouts: https://github.com/YACReader/yacreader/blob/350234b0991e79079281e4ee1ca95047410f7a9c/YACReader/main_window_viewer.cpp#L1040, https://github.com/YACReader/yacreader/blob/350234b0991e79079281e4ee1ca95047410f7a9c/YACReader/magnifying_glass.cpp#L228, https://github.com/YACReader/yacreader/blob/350234b0991e79079281e4ee1ca95047410f7a9c/YACReader/viewer.cpp#L580, https://github.com/YACReader/yacreader/blob/350234b0991e79079281e4ee1ca95047410f7a9c/YACReader/viewer.cpp#L655

Why are these custom events needed? If these shortcuts were implemented in this way by mistake or for some no longer valid reason, I would like to remove the custom event handlers and add missing `addAction()` and `connect()` calls. See the WIP diff in this pull request.

Replacing `orphanActions` with `this` is not necessary to make the shortcuts work, but I don't understand why `MainWindowViewer` cannot be a parent and would rather fix this (however benign) memory leak.


YACReader/main_window_viewer.cpp
```diff
@@ -1039,6 +1039,11 @@ void MainWindowViewer::disableActions()
 
 void MainWindowViewer::keyPressEvent(QKeyEvent *event)
 {
+    if (1) {
+        QMainWindow::keyPressEvent(event);
+        return;
+    }
+
```
selmf, line 1046: This is probably an old workaround to fix problems with multi key shortcuts, i.e. "CTRL+F" for fullscreen and other possible stuff. If you can make this work with all event modifier keys below there is no need for a custom keyPressEvent function at all, so the best approach is to simply remove it.

YACReader/main_window_viewer.cpp
```diff
@@ -1243,7 +1248,7 @@ void MainWindowViewer::processReset()
 void MainWindowViewer::setUpShortcutsManagement()
 {
     //actions holder
-    auto orphanActions = new QObject;
+    auto orphanActions = this;
```
selmf, line 1251:
```suggestion
    auto orphanActions = new QObject(this);
```
The variable name @luisangelsm used here suggest he had a reason to set this as a plain QObject. We can fix the memleak easily by setting "this" as the QObject's parent but we still need confirmation why he intended this as an orphaned object.

selmf (changes requested):
I've attache some comments and suggestions, but you should clarify the orphanActions mechanism with @luisangelsm

luisangelsm:
Some of the handlers are there to help with the events propagation. IIRC the magnifying glass wasn't getting the right events because parent widgets would eat them. The go to flow widget or the translator are also worth checking for possible events issues.

Others, like the `moveAction` method is there to help funneling  any shortcut to the default `QAbstractScrollArea` behavior.

The `orphanActions` bit is probably to avoid some propagation of events captured by `MainWindowViewer` but I don't remember the specifics.

I don't mind any changes that fix the keyboard layout problems as long as the behavior of all the actions doesn't change or gets broken. This will require some thorough testing before merging.

vedgy:
Both magnifying glass and `moveAction` shortcuts should be implementable with `QAction`s. If not, I'll leave these bits as is.

I am going to fix the shortcuts and the memory leak and do some testing before updating this pull request. If something breaks, I'll revert and document it for the benefit of future developers.

selmf:
@vedgy are you making progress on this? @luisangelsm would like to have the layout issue fixed in the next release if possible, so our interest in a mergeable version is high. Please let us know if you need more information or help otherwise.

vedgy:
I got distracted/absorbed by fixing Library bugs and put aside this PR. I didn't know you were particularly interested in it. Now I'll prioritize this over my unfinished Library fixes.

YACReader/main_window_viewer.cpp
```diff
@@ -1596,6 +1476,46 @@ void MainWindowViewer::getSiblingComics(QString path, QString currentComic)
     updatePrevNextActions(index > 0, index + 1 < list.count());
 }
 
+void MainWindowViewer::setActionsEnabled(bool enabled)
+{
+    //TODO enable goTo and showInfo (or update) when numPages emited
```
vedgy, line 1481: This pull request enables only the actions that used to be handled by `keyPressEvent()`s when `Render::numPages` => `Viewer::comicLoaded` signal is emitted. This old [merely moved in this PR] comment suggests that a couple more actions should be enabled at this signal rather than when they are now - when opening/loading/extracting a comic only begins. I wonder, perhaps *all* these actions should be enabled at the later time when the comic is completely loaded/extracted?

If not, then at least one call to `enableActions()` should be removed as redundant from `MainWindowViewer::open(QString path, qint64 comicId, qint64 libraryId)`, because it either fails or calls `MainWindowViewer::open(QString path, ComicDB &comic, QList<ComicDB> &siblings)`, which in turn also calls `enableActions()`.

YACReader/main_window_viewer.cpp
```diff
@@ -1596,6 +1476,46 @@ void MainWindowViewer::getSiblingComics(QString path, QString currentComic)
...
+
+void MainWindowViewer::setMglassActionsEnabled(bool enabled)
+{
+    for (auto *a : qAsConst(mglassActions))
+        a->setEnabled(enabled);
+}
+
+void MainWindowViewer::setLoadedComicActionsEnabled(bool enabled)
+{
+    for (auto *a : qAsConst(loadedComicActions))
```
vedgy, line 1515: [`qAsConst`](https://doc.qt.io/qt-5/qtglobal.html#qAsConst) used here and in `MainWindowViewer::setMglassActionsEnabled()` was introduced in Qt 5.7, so YACReader will require Qt 5.7+ when this PR is merged. It would be better to use standard [`std::as_const`](https://en.cppreference.com/w/cpp/utility/as_const) when we start requiring C++17.

YACReader/viewer.cpp
```diff
@@ -179,6 +185,7 @@ void Viewer::createConnections()
     connect(render, SIGNAL(numPages(unsigned int)), goToFlow, SLOT(setNumSlides(unsigned int)));
     connect(render, SIGNAL(numPages(unsigned int)), goToDialog, SLOT(setNumPages(unsigned int)));
     //connect(render,SIGNAL(numPages(unsigned int)),this,SLOT(updateInformation()));
+    connect(render, QOverload<unsigned int>::of(&Render::numPages), this, &Viewer::comicLoaded);
```
vedgy, line 188: `QOverload` was introduced in Qt 5.7, so YACReader will require Qt 5.7+ when this PR is merged. We can use more concise [`qOverload`](https://doc.qt.io/qt-5/qtglobal.html#qOverload) when we start requiring C++14 or C++17.

#197 disambiguates `Render::numPages` signal by renaming it to `Render::numPagesChanged`. Whichever of these two PRs is merged later, will have to be rebased and made compile.

YACReader/main_window_viewer.cpp
vedgy: I haven't noticed any issues with using `this` (`MainWindowViewer`) as the actions' parent. There was an issue when I tried `viewer->addAction()`: the actions appeared in `Viewer`'s context menu. So I ended up calling [`this->`]`addAction()` instead.

YACReader/main_window_viewer.cpp
vedgy: I have verified that the standard `QAction` shortcut mechanism supports all 4 modifier keys that were handled in `keyPressEvent()`s.

luisangelsm:
@vedgy do you have time for rebasing this one on top of develop?

vedgy:
I do have time, but not right now. Maybe in a few weeks.

YACReader/main_window_viewer.cpp
vedgy: No longer a problem because Qt 5.15 is the minimum supported version now. Switched to `std::as_const` as C++17 is now required.

YACReader/viewer.cpp
vedgy: No longer a problem because Qt 5.15 is the minimum supported version now. Switched to `qOverload` as C++17 is now required.

#197 will still have to be rebased (it has other conflicts now too).

vedgy:
Rebased as requested. Will rebase other pull requests too when you are ready to review them. Just ping me in a comment as you've done here.

YACReader/main_window_viewer.cpp
vedgy: @luisangelsm, what do you think about the issue described in this comment? This PR preserved the existing times when the actions were enabled/disabled. But since the implementation had been nonuniform, the difference might be unintended. The said difference is now reflected in the different places where `enableAcitons()` and `setLoadedComicActionsEnabled()` are invoked. I am wondering if all these actions should be handled uniformly. Or perhaps at least some actions should be moved between the two groups.

Issue #143, opened 2020-07-12
Add automatic code analysis to our CI


selmf:
We should use static code analysis to check for errors in our codebase and especially new commits and pull requests. There are several options available:

- Clang static analyzer
- Clazy static analyzer (this one is specifically written for Qt based code)
- Cppcheck
- Some proprietary services like Coverity that offer free checking to OpenSource projects

PR #193, opened 2021-02-04
Library: don't leak folderQueryResultProcessor at exit

vedgy:
`FolderQueryResultProcessor` has a `ConcurrentQueue` data member. The leak meant that the thread was not joined before exit.

### DIFF
--- a/YACReaderLibrary/db/folder_query_result_processor.cpp
+++ b/YACReaderLibrary/db/folder_query_result_processor.cpp
@@ -24,60 +24,62 @@ void YACReader::FolderQueryResultProcessor::createModelData(const YACReader::Sea
 {
     querySearchQueue.cancellPending();
 
-    QString connectionName = "";
-    {
-        QSqlDatabase db = DataBaseManagement::loadDatabase(model->getDatabase());
+    querySearchQueue.enqueue([=] {
+        QString connectionName = "";
+        {
+            QSqlDatabase db = DataBaseManagement::loadDatabase(model->getDatabase());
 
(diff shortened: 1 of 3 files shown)


vedgy:
`FolderQueryResultProcessor`'s `ConcurrentQueue querySearchQueue` data member is not *really* used. Therefore a thread runs all the time and does nothing useful. Should the data member be removed or should the bulk of [`FolderQueryResultProcessor::createModelData()`](https://github.com/YACReader/yacreader/blob/117b02fc7f789e2f3021de7a5be7d04088cb2371/YACReaderLibrary/db/folder_query_result_processor.cpp#L23) run in `querySearchQueue`'s thread like in [`ComicQueryResultProcesor::createModelData()`](https://github.com/YACReader/yacreader/blob/117b02fc7f789e2f3021de7a5be7d04088cb2371/YACReaderLibrary/db/comic_query_result_procesor.cpp#L28)?

luisangelsm:
Yes, it should be used, I removed the call momentarily to do some debugging and I committed the wrong code.

vedgy:
> Yes, it should be used, I removed the call momentarily to do some debugging and I committed the wrong code.

I could remove the data member myself. But actually using it requires better understanding of the code and knowing how to test the feature. So I'll leave this task to you.

luisangelsm:
Sure, I'll push it to this PR later. And thanks for the PR!

selmf:
I think we should look into adding some static code analyzing to our CI to catch these bugs before they are committed. See #143

luisangelsm:
@selmf yeah :)

PR #196, opened 2021-02-06
Fix some compiler warnings

vedgy:

### DIFF
--- a/YACReader/main_window_viewer.cpp
+++ b/YACReader/main_window_viewer.cpp
@@ -948,7 +948,7 @@ void MainWindowViewer::openFolderFromPath(QString pathDir, QString atFileName)
     d.setSorting(QDir::Name | QDir::IgnoreCase | QDir::LocaleAware);
     QStringList list = d.entryList();
 
-    qSort(list.begin(), list.end(), naturalSortLessThanCI);
+    std::sort(list.begin(), list.end(), naturalSortLessThanCI);
     int i = 0;
     foreach (QString path, list) {
         if (path.endsWith(atFileName))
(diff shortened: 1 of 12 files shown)


YACReaderLibrary/server/controllers/v2/pagecontroller_v2.cpp
```diff
@@ -33,7 +31,6 @@ void PageControllerV2::service(HttpRequest &request, HttpResponse &response)
     bool remote = path.endsWith("remote");
 
     QStringList pathElements = path.split('/');
-    QString libraryName = DBHelper::getLibraryName(pathElements.at(2).toInt());
```
vedgy, line 36: This and the identical change in `PageController::service()` are the ones I'm not sure about. If the call to `DBHelper::getLibraryName()` is useful here, either the variable assignment should be removed or an explaining comment should be added.

selmf:
From a first glance, this is a bit of a mixed bag. A good amount of these are no-brainers and can be easily merged but there are some commit thrown in (like the pagecontroller) that need closer examination which I can't provide given that I don't use YACReader for iOS.
@vedgy can I trouble you to put the commits that need examination into a separate PR? I'm currently working on something that requires the removal of deprecated api usage and it would be stupid to duplicate your work.

vedgy:
I think the only risky commit here is the `PageController` one. I can move that in a separate pull request.

> I'm currently working on something that requires the removal of deprecated api usage and it would be stupid to duplicate your work.

I've also ported away from some deprecated Qt API in #206, so you may want to look at it as well.

vedgy:
Split the `PageController` commit into #231.

PR #197, opened 2021-02-06
index/numPages/numSlides: unsigned => int

vedgy:
This change gets rid of a few GCC's *-Wsign-compare* warnings and several explicit casts because of different signedness.

Make several getters const member functions rather than slots.

Rename several signals from `numPages()` to `numPagesChanged()` to simplify using function pointer-based signal-slot connection syntax.

### DIFF
--- a/YACReader/goto_dialog.cpp
+++ b/YACReader/goto_dialog.cpp
@@ -68,7 +68,7 @@ void GoToDialog::goTo()
     close();
 }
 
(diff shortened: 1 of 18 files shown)


vedgy:
I've cleaned up `Render` code somewhat in https://github.com/vedgy/yacreader/commits/clean-up-render based on this branch. I'll create a separate pull request after this one is merged to avoid resolving merge conflicts repeatedly.

PR #200, opened 2021-02-06
Fix a typo in ComicQueryResultProcessor class name

vedgy:
See also the second commit message.

### DIFF
--- a/YACReaderLibrary/db/comic_query_result_processor.h
+++ b/YACReaderLibrary/db/comic_query_result_processor.h
@@ -1,5 +1,5 @@
-#ifndef COMIC_QUERY_RESULT_PROCESOR_H
-#define COMIC_QUERY_RESULT_PROCESOR_H
+#ifndef COMIC_QUERY_RESULT_PROCESSOR_H
+#define COMIC_QUERY_RESULT_PROCESSOR_H
 
 #include <QtCore>
 #include <QSqlQuery>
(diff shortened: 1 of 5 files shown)


YACReaderLibrary/db/comic_query_result_processor.h
```diff
@@ -25,7 +25,7 @@ public slots:
 private:
     ConcurrentQueue querySearchQueue;
 
-    QList<ComicItem *> *modelData(QSqlQuery &sqlquery);
+    static QList<ComicItem *> *modelData(QSqlQuery &sqlquery);
```
luisangelsm, line 28: Is doing this change worth it? I usually don't bother unless the method will be used in a static way, the performance gain in this case is negligible. So what's the motivation?

The typo fix is so straightforward that I don't understand the need of adding anything else to the PR.

vedgy: Just tried to find a C++ guideline that recommends making member functions static if possible, but surprisingly failed :)

My primary motivation is to improve readability of the code, which is more important than speeding up the process of writing code as it is usually read much more often than modified. When a static function is called, the reader of the code knows for sure that it won't access any data members. This is similar but stronger than a `const` qualifier, which *is* recommended by most modern C++ guidelines. For example: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconst-fct.

In a multithreading context a `static` member function can be run concurrently with a *non*-`const` member function w/o synchronization. While a `const` member function can be run concurrently only with `const` member functions w/o synchronization.

PR #206, opened 2021-02-10
Fix some Qt deprecation warnings, memory leaks; clean up related code

vedgy:

### DIFF
--- a/YACReader/magnifying_glass.cpp
+++ b/YACReader/magnifying_glass.cpp
@@ -141,31 +141,34 @@ void MagnifyingGlass::updateImage()
 }
 void MagnifyingGlass::wheelEvent(QWheelEvent *event)
 {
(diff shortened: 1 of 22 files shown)


YACReader/magnifying_glass.cpp
```diff
@@ -141,31 +141,34 @@ void MagnifyingGlass::updateImage()
 }
 void MagnifyingGlass::wheelEvent(QWheelEvent *event)
 {
+    // TODO: consider handling horizontal scrolling differently.
+    const auto delta = event->angleDelta().x() + event->angleDelta().y();
+    const bool grow = delta < 0;
```
vedgy, line 146: The zoom in/out direction of the Magnifying glass is opposite to the direction in other applications. Is this intentional or an oversight that should be fixed by initializing `grow` with `delta > 0`?

PR #208, opened 2021-02-11
GridComicsView: highlight selected and hovered-over covers differently

vedgy:
When the selected and the mouse-hovered-over cover have the same border colors, they cannot be distinguished. The difference is important because the selected/current comic is opened when the _Open current comic_ shortcut is activated. The identical highlighting is especially confusing during keyboard navigation.

### DIFF
--- a/YACReaderLibrary/qml/GridComicsView.qml
+++ b/YACReaderLibrary/qml/GridComicsView.qml
@@ -130,7 +130,7 @@ Rectangle {
                         rightMargin  : commonBorder ? -commonBorderWidth : -rBorderwidth
                     }
 
-                    border.color: (Qt.platform.os === "osx") ? selectedBorderColor : "#ffcc00"
+                    border.color: (dummyValue || !dummyValue) && comicsSelectionHelper.isSelectedIndex(index) ? (Qt.platform.os === "osx" ? selectedBorderColor : "#ffcc00") : "#996600"
                     border.width: 3
 
                     opacity: (dummyValue || !dummyValue) && (comicsSelectionHelper.isSelectedIndex(index) || mouseArea.containsMouse) ? 1 : 0


luisangelsm:
I tested this time ago and I don't link the different color approach...

I've tried changing the scale of the whole `cell`

`scale: mouseArea.containsMouse ? 1.025 : 1`

But maybe it is too subtle...

vedgy:
I don't think changing the scale is too subtle. When the mouse pointer hovers over something, the UI doesn't even have to react (most of the time it does not). The enlargement of the hovered-over comics cover is easily noticeable while it happens. Afterwards, there is no need to highlight which cover the mouse cursor hovers over because it does not matter for anything other than a mouse click. The following patch works for me:

```diff
--- a/YACReaderLibrary/qml/GridComicsView.qml
+++ b/YACReaderLibrary/qml/GridComicsView.qml
@@ -60,6 +60,7 @@ Rectangle {
             width: grid.cellWidth
             height: grid.cellHeight
             color: "#00000000"
+            scale: mouseArea.containsMouse ? 1.025 : 1
 
             DropShadow {
                 anchors.fill: realCell
@@ -133,7 +134,7 @@ Rectangle {
                     border.color: (Qt.platform.os === "osx") ? selectedBorderColor : "#ffcc00"
                     border.width: 3
 
-                    opacity: (dummyValue || !dummyValue) && (comicsSelectionHelper.isSelectedIndex(index) || mouseArea.containsMouse) ? 1 : 0
+                    opacity: (dummyValue || !dummyValue) && comicsSelectionHelper.isSelectedIndex(index) ? 1 : 0
 
                     Behavior on opacity {
                         NumberAnimation { duration: 300 }
```

luisangelsm:
`scale` is used now in develop to differentiate selection vs mouse over states

PR #209, opened 2021-02-11
Add focus shortcuts

vedgy:


luisangelsm:
Instead of disabling this in macos, we need to add a method to `YACReaderMacOSXSearchLineEdit` to set the focus, `YACReaderMacOSXSearchLineEdit` uses a `NSTextField` internally and we can call `[field becomeFirstResponder]`.

vedgy:
I could try to read the docs and implement the Mac part, but I won't be able to test it as I don't have access to Mac.

luisangelsm:
It will be enough if you remove those `#ifndef` and add an empty method/slot to `YACReaderMacOSXSearchLineEdit` to gain focus, I'll commit the tested macos part after that.

vedgy:
The first force-push is a rebase on the current *develop*. The second force-push contains the actual changes to the previous version, so reviewers can use the second *force-pushed* link to view the interesting diff. In addition to the requested macOS change, I have removed unused `ComicsViewTransition::sizeHint()` and added a `TODO` comment (see the third commit message).

luisangelsm:
Tested on macOS, working as expected.

vedgy:
By the way, I have a WIP branch where pressing *Return*/*Enter* transfers focus from the search edit to the comics view. Plus when a folder or a reading list is activated, comics view gets the focus too. In the process of implementing these features I've found numerous Library bugs, most of which are fixed in #219 and the branch mentioned in a comment to #219.

PR #212, opened 2021-02-12
Add CONFIG += silent to build to reduce log pollution

selmf:
Qmake by default creates lots of unnecessary information during builds which makes it hard to extract warnings and errors from the log.

This PR adds the undocumented silent parameter to make it less talkactive.


vedgy:
The build output contains much less useless text now. The output became more like CMake's. When I merged this branch into my working branch, the full build log size shrunk from 1.6 MiB to 340 KiB. This obscure parameter should be on by default in qmake :)

PR #214, opened 2021-02-14
Library: don't leak DB model objects

vedgy:

### DIFF
--- a/YACReaderLibrary/db/comic_model.h
+++ b/YACReaderLibrary/db/comic_model.h
@@ -64,8 +64,8 @@ public:
     };
 
 public:
-    ComicModel(QObject *parent = 0);
-    ComicModel(QSqlQuery &sqlquery, QObject *parent = 0);
+    explicit ComicModel(QObject *parent = nullptr);
+    explicit ComicModel(QSqlQuery &sqlquery, QObject *parent = nullptr);
     ~ComicModel() override;
 
     QVariant data(const QModelIndex &index, int role) const override;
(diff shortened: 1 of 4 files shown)


selmf:
These are not leaks. If I am not mistaken, the DB model objects are instantiated only once during the app lifetime and they do not leak memory during usage. So these are actually missing cleanups or, if you insist on calling it so, static leaks.

These are still bugs and they need to be fixed. However, the commit messages should reflect what is actually being done (add the missing cleanups) and they should be written in a technical way (no first person references).

If you adjust the commits accordingly, I can merge this PR.

vedgy:
> These are not leaks. If I am not mistaken, the DB model objects are instantiated only once during the app lifetime and they do not leak memory during usage. So these are actually missing cleanups or, if you insist on calling it so, static leaks.

You are right that *currently* these objects are created once per application run. This might change in the future if the program creates multiple instances of the class that should own them or if their initialization/"ownership" is delegated to another, non single-instance class.

In any case, never destroying the objects is still a leak in my opinion, regardless of their lifetime. Turns out not everyone agrees about what is and is not a leak. The top answer to [this question](https://stackoverflow.com/questions/273209/are-memory-leaks-ever-ok) agrees with me. Many other answers and comments match your terminology.

The question itself mentions one disadvantage of such "benign" leaks: the difficulty of distinguishing them from more consequential resource leaks in a leak detection tool's output. I'd argue that not treating never destroyed objects as serious issues increases the likelihood of more harmful resource leaks (and not just because of the obstacle to memory leak detection tool use). I agree with the top Stack Overflow answer's arguments too.

That said, I agree with you that the "benign" nature of the leaks should be described in the commit message.

> ... and they should be written in a technical way (no first person references).

I've never seen such a guideline for the commit message body anywhere. Just tried to find something like this and failed. Is this your personal preference or a common practice that I missed? Could you share a link to some article about this? As a counterargument, consider this recent commit message in *git* itself: https://git.kernel.org/pub/scm/git/git.git/commit/?id=bbabaad29823890f328086c6f11b3dddc118adb8

selmf:
There is no need to defend yourself for wanting to fix these bugs. It is pretty clear what they are and that they have potential for troubles in the future if left alone.

Using a technical writing style for the commit messages is not a personal preference of mine but a common practice. For most projects this is likely an unwritten rule that most people follow by instinct because they are taught to do it this way.

If you want some guidelines, Chris Beam has some good tips: 

[How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit)

Note that I am not a stickler to rules and that I acknowledge that there are times where using first person or non-technical writing in a commit message makes perfectly sense. For example, this might be the case if you want to add information that requires expression of a personal opinion or you want to include a suggestion on how to further continue from this point.
However, if you find yourself wanting to do that you should always check if it would not be a good idea to add this information to the PR directly so that it can easily be viewed and discussed.

selmf:
@luisangelsm This looks good to me. Any objections to a merge?

vedgy:
I've read the guidelines at your link a long time ago. They don't advocate avoiding first person references in the commit message **body**. I don't see what is non-technical about the former commit body:
> I have verified that these objects were never destroyed before this commit by printing debug messages in their destructors.

That sentence seems more precise to me than the current impersonal wording. And precision is important in technical language. But I've rephrased the commit message hopefully to your liking :)

luisangelsm (approved):

PR #222, opened 2021-02-27
fix: Replace QtScript with QJson*

luisangelsm:
QtScript is deprecated and this was needed to start supporting Qt6

This has been cherry-picked and fixed from: https://github.com/u19809/yacreader/tree/No-QtScript-and-Warnings all the code in the branch assumed that JSON object were all maps, but there are a bunch of objects coming from Comic Vine that are Arrays, so the hole thing was pretty much broken.


selmf:
Looks like the Windows installer is still trying to include QtScript

PR #225, opened 2021-03-03
Don't crash during a slow (>= 60 seconds) startup on X11

vedgy:
See the added source code comments and the commit message for details.

The patch I used for benchmarking (stupid GitHub won't let me attach a patch => had to zip it): [0001-Benchmark-QCoreApplication-processEvents-workaround.patch.zip](https://github.com/YACReader/yacreader/files/6075476/0001-Benchmark-QCoreApplication-processEvents-workaround.patch.zip).

### DIFF
--- a/YACReader/main.cpp
+++ b/YACReader/main.cpp
@@ -98,6 +98,12 @@ int main(int argc, char *argv[])
 #else
     QApplication app(argc, argv);
 #endif
+    // Prevent SIGPIPE, then "ICE default IO error handler doing an exit(), pid = <PID>, errno = 32"
+    // crash on X11 when the first event loop starts at least 60 seconds after a Qt application
+    // launch. This can happen during a Debug launch of YACReader from Qt Creator if Qt debug
+    // symbols are installed in the system or a breakpoint is hit before any event loop is entered.
+    // This is a workaround for QTBUG-58709.
+    QCoreApplication::processEvents();
 
 #ifdef FORCE_ANGLE
     app.setAttribute(Qt::AA_UseOpenGLES);
--- a/YACReaderLibrary/main.cpp
+++ b/YACReaderLibrary/main.cpp
@@ -120,6 +120,11 @@ int main(int argc, char **argv)
 {
     qInstallMessageHandler(messageHandler);
     QApplication app(argc, argv);
+    // Prevent SIGPIPE, then "ICE default IO error handler doing an exit(), pid = <PID>, errno = 32"
+    // crash on X11 when the first event loop starts at least 60 seconds after a Qt application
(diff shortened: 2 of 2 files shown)


selmf:
Is this a good idea? If I understand your testing process correctly, the crash you want to fix here occurs in a highly artificial scenario (forced interruption/delay of main event loop) and is only needed for some very specific testing scenarios. So the chances of this happening during a normal startup are very slim and the usefulness of enforcing process events are doubtful.
 
If it really is needed, it should be contained to the specific scenario it was written for, i.e. debug builds only and possibly X11 only too if it is irrelevant for the other systems.

vedgy:
I have been hit by this crash many times over the last month while debugging Reader and Library, and not in *highly* artificial scenarios. Though of course end users won't encounter it unless their computers are extremely slow.

The actual scenarios:
1. Sometimes a developer uses a debugger to step through code that runs at the application's startup. If the `Q(Core|)Application::exec()` call in `main()` is thus delayed by more than a minute, the debugging session is interrupted by the crash.
2. I have recently rebuilt my system *qt5-base* package by applying diffs from https://github.com/qutebrowser/qt-debug-pkgbuild/tree/master/qt5-base to be able to step into Qt sources while debugging. Turns out that the already slow YACReader debug session startup becomes much slower when Qt debug symbols are available - now it always takes more than 60 seconds on my system. Thus attaching to an already running Reader process became the only alternative to this workaround. The attaching requires changing system *ptrace_scope* permissions, is less convenient and does not allow to debug the application startup.

Restricting the workaround to Debug builds and/or X11 has a potential to introduce inconsistencies between platforms and build modes. Some bugs may manifest themselves with the event processing, but not without, or the other way around. This can easily become a maintenance burden and a source of Release-only or platform-specific bugs. I think it is best to apply and try it out on all supported platforms. Hopefully it won't cause any issues and can remain unconditional. If you don't want to risk introducing a bug into the upcoming YACReader release, this workaround can wait until after the new version is released. Then we will have an entire release cycle to notice regressions.

vedgy:
Just found the upstream bug report: [QTBUG-58709](https://bugreports.qt.io/browse/QTBUG-58709). Linked it with the Stack Overflow workaround and replaced other links in this PR's commit message with the Qt bug ID.

selmf:
I am still undecided on the best way to proceed here and I will likely need to do some more reflection and some tests of my own before reaching a decision.

As a general rule, if there is a QTBUG that we need to implement a workaround for we always include a comment with the bug id so we can easily identify and check it at a later point in development. So please add

`// Workaround for QTBUG-58709`

or something along the lines to the relevant sections so we can find it when grepping for QTBUG :)

PR #228, opened 2021-03-06
Feature: make comic vine dialog more keyboard friendly

luisangelsm:


config.pri
```diff
@@ -14,11 +14,10 @@ QT_VER_MAJ = $$member(QT_VERSION, 0)
 QT_VER_MIN = $$member(QT_VERSION, 1)
 
 lessThan(QT_VER_MAJ, 5) {
-error(YACReader requires Qt 5 or newer but Qt $$[QT_VERSION] was detected.)
-  }
 lessThan(QT_VER_MIN, 6):!CONFIG(no_opengl) {
   error ("You need at least Qt 5.6 to compile YACReader or YACReaderLibrary.")
   }
+}
```
vedgy, line 20: I think this change is incorrect. The `lessThan(QT_VER_MAJ, 5)` should be restored to reject Qt 4. And the `lessThan(QT_VER_MIN, 6)` should be enclosed in a separate `equals(QT_VER_MAJ, 5)` block.

luisangelsm: Absolutely

config.pri
```diff
@@ -14,11 +14,14 @@ QT_VER_MAJ = $$member(QT_VERSION, 0)
 QT_VER_MIN = $$member(QT_VERSION, 1)
 
 lessThan(QT_VER_MAJ, 5) {
-error(YACReader requires Qt 5 or newer but Qt $$[QT_VERSION] was detected.)
-  }
+  error ("You need at least Qt 5.6 to compile YACReader or YACReaderLibrary.")
+}
+
+equals(QT_VER_MAJ, 5) {
 lessThan(QT_VER_MIN, 6):!CONFIG(no_opengl) {
```
vedgy, line 21: This `:!CONFIG(no_opengl)` subcondition suggests that YACReader can be built with Qt5 version < 5.6 if `no_opengl` config variable is set. If so, then the error message replacement in the `lessThan(QT_VER_MAJ, 5)` block is misleading and the old error message should probably be restored. If Qt < 5.6 is not supported at all, perhaps `:!CONFIG(no_opengl)` should be removed.

selmf: I think this was added for some users stuck on older versions of Qt on Debian or Mepis Linux. The only part of YACReader requiring Qt 5.6 at that time was the cover flow, so you could still build with older versions by disabling the opengl widgets. That likely is no longer the case.

I think we should use the occasion and increase the minimum version requirements for Qt. There are probably not many people left using these ancient Qt versions to build YACReader and the LTS versions of Debian and Ubuntu that are the main target for such legacy builds already have newer versions packaged. Given that our current code base uses functionality from Qt 5.8 and Ubuntu 18.04 ships with Qt 5.9 I suggest using that version as our new minimum target.

selmf:
The Linux build will likely fail as our CI still runs an older Ubuntu version. This is not hard to fix, but a few workarounds we no longer need need to be removed. I'll take care of it tomorrow.

luisangelsm:
> 
> 
> The Linux build will likely fail as our CI still runs an older Ubuntu version. This is not hard to fix, but a few workarounds we no longer need need to be removed. I'll take care of it tomorrow.

Then I'll remove the config.pri changes from this PR and we can do it in a separate PR

selmf:
Yeah, I just checked. Our Linux CI is still using Qt 5.6.1 - which means we can still build using that version. Not hard to change this, we just need to change the VM image to Ubuntu 18.04 and remove the external PPA package source for the Qt packages as we're switching to the shipped packages.

Still a good idea to switch to Qt 5.9 as a base as this is much closer to what we're actually using to test and develop. Anybody wanting to build with 5.6 still can do so, but we don't have to support it :)

vedgy:
I'm all for increasing minimum supported Qt version in order to be able to use things like [qOverload](https://doc.qt.io/qt-5/qtglobal.html#qOverload) and other features introduced between versions 5.6 and 5.9. Ubuntu 16.04 is supported only until April 30th, 2021. Could we perhaps increase the C++ standard version to C++17 too? Ubuntu 18.04's GCC 7.4 supports almost all C++17 features: https://gcc.gnu.org/projects/cxx-status.html#cxx17 and https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2017. If we don't have to support old Clang and MSVC versions, C++17 should be almost completely supported on Mac and Windows too. The newer standard makes development more fun and motivating.

selmf:
@vedgy Luis and I discussed this in the past and the possibility definitely is on the table, but I'd suggest we use a separate issue for a separate discussion in this case. This is a merged PR and if we continue here, the information will be hard to find and access for other interested individuals.

PR #231, opened 2021-03-07
PageController*::service: remove unused libraryName variable

vedgy:
Hopefully the code does not rely on the side effects of the removed call to `DBHelper::getLibraryName()`.

If the call to `DBHelper::getLibraryName()` is useful, either the variable assignment should be removed or an explaining comment should be added.


luisangelsm:
This is in other controllers too, it shouldn't be necessary, so I think we should remove it in all the controllers that don't use the library name and then test the server just in case.

vedgy:
> we should remove it in all the controllers that don't use the library name

Done. All other controllers use this variable. Looks like there is lots of code duplication in the controllers.

luisangelsm:
Looking good, thanks.

Issue #230, opened 2021-03-07
C++17 in YACReader


selmf:
> I'm all for increasing minimum supported Qt version in order to be able to use things like qOverload and other features introduced between versions 5.6 and 5.9. Ubuntu 16.04 is supported only until April 30th, 2021. Could we perhaps increase the C++ standard version to C++17 too? Ubuntu 18.04's GCC 7.4 supports almost all C++17 features: https://gcc.gnu.org/projects/cxx-status.html#cxx17 and https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2017. If we don't have to support old Clang and MSVC versions, C++17 should be almost completely supported on Mac and Windows too. The newer standard makes development more fun and motivating.
> - @vedgy in #228

> About c++17, since Qt6 adopts c++17 and we want to move in that direction I think it is reasonable to start using it too.
> - @luisangelsm in #201 

I have copied these here so the discussion has a proper place.

IMHO moving to c++17 is the obvious next step, but we should consider when to do it. Personally I would prefer to get our code into a better shape first, so we've got our backs covered before marching forward. I'm not thinking about big rewrites here but more along the lines of elimination of deprecated code, tool assisted refactoring for stuff like old style SIGNAL SLOT connects and the like.

There is also the question of build chain support. Qmake 5.9 does not support CONFIG+=c++17, so we will have to add it to the compile flags manually. Porting or testing to Qt6 will make little sense before Qt 6.2 is released since Qt6 does not provide all modules we require yet, so it is unlikely we will do that for YACReader 9.8.

This would make C++17 a good milestone for YACReader 9.9, at which point we also could re-asses the Ubuntu situation regarding full C++17 support.

Until then we could treat this like we have done for older Qt versions in the past - use modern c++17 functions where appropriate, provide fallback solutions where necessary and restrict newer functionality to newer systems if that is too complicated.

luisangelsm:
I think we can use `CONFIG += c++1z` to enable c++17 in Qt5.9

vedgy:
> I think we can use CONFIG += c++1z to enable c++17 in Qt5.9

That would be great. But if it [doesn't work](https://doc.qt.io/archives/qt-5.9/qmake-variable-reference.html#config), we can at least enable c++14 for now.

> This would make C++17 a good milestone for YACReader 9.9, at which point we also could re-asses the Ubuntu situation regarding full C++17 support.

[...]

PR #232, opened 2021-03-07
Switch to C++17 and Qt 5.9 as a minimum supported version

vedgy:
See #230.

### DIFF
--- a/config.pri
+++ b/config.pri
@@ -2,7 +2,7 @@
 # default values if they're not set on build time
 # for a more detailed description, see INSTALL.TXT
 
-CONFIG += c++11
+CONFIG += c++1z
 
 unix:QMAKE_CXXFLAGS_RELEASE += -DNDEBUG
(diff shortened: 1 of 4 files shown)


config.pri
```diff
@@ -2,7 +2,7 @@
 # default values if they're not set on build time
 # for a more detailed description, see INSTALL.TXT
 
-CONFIG += c++11
+CONFIG += c++1z
```
vedgy, line 5: This line should be moved below the Qt version checks to avoid a confusing no-such-CONFIG-value error when the Qt version is too old.

vedgy: We can use the following code from *qtcreator.pri* for better compatibility:
```qmake
# include c++1z as an alias for c++17 for compatibility for older Qt versions
CONFIG += c++17 c++1z
```

luisangelsm:
Done as part of the migration to Qt6

PR #234, opened 2021-03-09
Translations/add missing forum translations

selmf:
This PR adds the missing translations for Chinese and Italian from the YACReader forum to our build and provides a much-needed update of our translation source .ts files, which have been synced with the community .ts files and updated against our source code.

For future updates, a CI solution would be preferable.


vedgy:
Ideally *.qm* files should not be stored in the git repository at all. The CI can run *lrelease* and generate them from *.ts* sources for each build.

Issue #73, opened 2019-08-23
Migrate to CMake


selmf:
YACReader currently uses qmake as build system, but this is showing its limits. We really need some proper dependency management, modularization and generally could use some improvements to reduce our build time and better handle our build targets.
While some of this probably could be done using qmake many 'advanced' Qt projects already use CMake for this purpose and even Qt upstream is slowly deprecating qmake in favor of it.

As this will mean that we'll be more or less rewriting our whole build system I propose to slowly introduce the new build system alongside our old build process and to start with the library server, as its build process will benefit the most from having better dependency detection and its code footprint is lower than the reader and the library targets.

capecrusader-121:
@selmf  Any updates on when this will be accomplished i am trying to snap yacreader but unfortunately the latest snap core22 containing Qt-5-15+ doesnt have a qmake plugin and as such it creates a roadblock to get the yacreader snaps published.

capecrusader-121:
@selmf  An experimental branch will be great, cmake works well with core22.

luisangelsm:
The build system has been migrated to cmake.

PR #237, opened 2021-03-12
Build/qt 5.9

selmf:
This PR improves our Qt version check macros for Qt 5 and fixes a false positive in the check that triggered when attempting to build using Qt 6.

Additionally, it increases the minimum build requirement to Qt 5.9 and introduces Qt deprecation macros as they are standard with every new Qt project.

This means that with the merge of this PR all use of Qt functions deprecated with 5.9 or earlier will lead to a build failure and newer Qt versions will display warnings when using functions deprecated after 5.9. This will allow us to defend against such functions sneaking back in and to start migrating away from functions deprecated in later versions. This is also a necessary step for future Qt 6 compatibility.

### DIFF
--- a/config.pri
+++ b/config.pri
@@ -8,17 +8,36 @@ unix:QMAKE_CXXFLAGS_RELEASE += -DNDEBUG
 win32:QMAKE_CXXFLAGS_RELEASE += /DNDEBUG
 
 # check Qt version
-QT_VERSION = $$[QT_VERSION]
-QT_VERSION = $$split(QT_VERSION, ".")
-QT_VER_MAJ = $$member(QT_VERSION, 0)
-QT_VER_MIN = $$member(QT_VERSION, 1)
+defineTest(minQtVersion) {
+  maj = $$1
+  min = $$2
+  patch = $$3
+  isEqual(QT_MAJOR_VERSION, $$maj) {
+    isEqual(QT_MINOR_VERSION, $$min) {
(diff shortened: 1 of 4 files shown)


config.pri
```diff
@@ -8,17 +8,39 @@ unix:QMAKE_CXXFLAGS_RELEASE += -DNDEBUG
...
-QT_VERSION = $$[QT_VERSION]
-QT_VERSION = $$split(QT_VERSION, ".")
-QT_VER_MAJ = $$member(QT_VERSION, 0)
-QT_VER_MIN = $$member(QT_VERSION, 1)
-
-lessThan(QT_VER_MAJ, 5) {
-error(YACReader requires Qt 5 or newer but Qt $$[QT_VERSION] was detected.)
+defineTest(minQtVersion) {
+  maj = $$1
+  min = $$2
```
vedgy, line 13: I was about to suggest renaming these variables to clearer `major` and `minor` until I learned that this function definition is copied from *qtcreator.pri*. Now I agree that the names should stay abbreviated to facilitate updating from upstream.

selmf: My hope is that we will have finished #73 before having to update this

selmf:
OK, seems like we are up to date with Qt 5.9. @luisangelsm I had to remove dead code specific to macOS and the custom treeview. Could you please have a short look at the last commit of this PR?

luisangelsm:
@selmf yeah, I think it is totally safe to remove that bit

Issue #201, opened 2021-02-07
ConcurrentQueue::cancellPending() modifies _queue while the wrong mutex is locked


vedgy:
`ConcurrentQueue::_queue` is protected by `queueMutex` everywhere but in [`cancellPending()`](https://github.com/YACReader/yacreader/blob/3a9bec55f8f882cf5acd1fb8ccb6eafbd7bd022d/common/concurrent_queue.h#L46).

[`ConcurrentQueue::waitAll()`](https://github.com/YACReader/yacreader/blob/3a9bec55f8f882cf5acd1fb8ccb6eafbd7bd022d/common/concurrent_queue.h#L53) is currently unused. Is `waitAll()` likely to be used somewhere soon? If not, it could be removed along with `jobsLeft`, `_waitVar` and `jobsLeftMutex`. This would greatly simplify the class. If a more feature-full concurrent queue is needed in the future, it might be a good idea to reuse some tested and tried implementation, such as [`QtConcurrent::run`](https://doc.qt.io/qt-5/qtconcurrentrun.html) (possibly with a dedicated thread pool).

luisangelsm:
Fixed [here](https://github.com/YACReader/yacreader/commit/4c93c70de6249678f77e12f94a12f1a7e2b2eabd).

Yes, the class as it is will be used in the future, I have a few things in mind that I would like to be pure c++.

vedgy:
Not really fixed. `jobsLeft` is also modified in this function, so `jobsLeftMutex` must also be locked. Possibly in separate code blocks to avoid locking both mutexes at the same time and reduce the risk of a deadlock. On the other hand, there might be reason to update both data members in a single transaction (while both mutexes are locked). More careful analysis is needed to tell for sure. Also note that there might be other bugs in the current complex code, especially in its currently unused parts.

Maybe it would be better to use an existing implementation such as `QtConcurrent` or https://github.com/cameron314/concurrentqueue. It could be faster too, especially a lock-free queue.

luisangelsm:
Ok, now.

As I said I would like to have some pure c++ solution to use in code that I want to keep isolated from Qt in the future.

vedgy:
[...]

PR #239, opened 2021-03-18
Test, fix, improve, optimize ConcurrentQueue

vedgy:

I figured that *concurrent_queue_test* does not require any sophisticated testing framework features. Since Qt Test works just fine for this purpose, I decided not to try out *Catch2* (as discussed in #201) for now.

Downsides of setting up a *Catch2 v3* development version dependency right now:
1. The development version is not yet very stable or well-documented. So YACReader's testing code might require frequent revisions when *Catch2* is updated.
2. *Catch2* is built with CMake, which is the build system YACReader is going to adopt in the future (see #73). But until CMake is actually adopted, mixing qmake and CMake can be unwieldy and inconvenient.
3. YACReader code is generally poorly prepared for testing, so I don't expect many tests to be written in the nearest future. The refactoring necessary to improve testability can take a lot of time. Thus a more convenient testing framework is unlikely to be highly beneficial soon.
4. Perhaps by the time YACReader code is ready for extensive testing, another testing framework would become more suitable for this purpose.
5. Qt Test is part of Qt Base, is very stable and does not require additional dependencies. In the end, Qt Test can be used interchangeably with a more advanced testing framework, whichever is more convenient for each particular testing task.

### DIFF
--- a/azure-pipelines-windows-template.yml
+++ b/azure-pipelines-windows-template.yml
@@ -34,6 +34,11 @@ jobs:
       nmake
     displayName: 'Build'
   - script: |
+      call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\${{ parameters.vc_vars }}"
+      set PATH=C:\Qt\${{ parameters.qt_version }}\${{ parameters.qt_spec }}\bin;%PATH%
+      nmake check TESTARGS="-maxwarnings 100000"
+    displayName: 'Run tests'
+  - script: |
       set PATH=C:\Qt\${{ parameters.qt_version }}\${{ parameters.qt_spec }}\bin;%PATH%
       cd $(Build.SourcesDirectory)\ci\win
       .\create_installer.cmd ${{ parameters.architecture }} 7z $(Build.BuildNumber)
--- a/common/concurrent_queue.cpp
+++ b/common/concurrent_queue.cpp
@@ -1,0 +1,127 @@
+#include "concurrent_queue.h"
+
+#include <cassert>
+#include <cstddef>
+#include <mutex>
+#include <utility>
+
+using namespace YACReader;
+
+ConcurrentQueue::ConcurrentQueue(std::size_t threadCount)
+{
+    threads.reserve(threadCount);
+    for (; threadCount != 0; --threadCount)
+        threads.emplace_back(&ConcurrentQueue::nextJob, this);
+}
+
+ConcurrentQueue::~ConcurrentQueue()
+{
+    joinAll();
+}
+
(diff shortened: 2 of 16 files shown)


azure-pipelines-windows-template.yml
```diff
@@ -32,6 +32,11 @@ jobs:
       qmake CONFIG+="7zip" %DEFINES_VAR%
       nmake
     displayName: 'Build'
+  - script: |
+      call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\${{ parameters.vc_vars }}"
+      set PATH=C:\Qt\${{ parameters.qt_version }}\${{ parameters.qt_spec }}\bin;%PATH%
+      nmake check TESTARGS="-maxwarnings 100000"
+    displayName: 'Run tests'
```
vedgy, line 39: Unfortunately, the test output on Windows is absent from [the CI build logs](https://dev.azure.com/luisangelsm/YACReader/_build/results?buildId=441&view=logs&jobId=05ed95bb-1786-5052-b294-220ffba395f2&j=b9e62f99-1a98-5ed7-01d2-f4794231ed79&t=c8064208-5108-5c88-7a6d-4ac7b3cf2e15): 

```cmd
"C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.28.29910\bin\HostX86\x86\nmake.exe" -f Makefile.Release check

Microsoft (R) Program Maintenance Utility Version 14.28.29910.0
Copyright (C) Microsoft Corporation.  All rights reserved.

	call D:\a\1\s\tests\concurrent_queue_test\target_wrapper.bat  release\concurrent_queue_test.exe -maxwarnings 100000
Finishing: Run tests
```

I suppose either `call` or `target_wrapper.bat` suppresses `concurrent_queue_test.exe`'s output. I looked for a fix but didn't find one.

common/concurrent_queue.cpp
```diff
@@ -0,0 +1,107 @@
...
+void ConcurrentQueue::enqueue(Job job)
+{
+    {
+        std::lock_guard<std::mutex> lock(jobsLeftMutex);
+        ++jobsLeft;
+    }
+
+    {
+        std::lock_guard<std::mutex> lock(queueMutex);
+        _queue.emplace(std::move(job));
```
vedgy, line 40: According to Sean Parent, `std::list` should be used instead of `std::queue` to minimize the locking time with  [`std::list::splice`](https://en.cppreference.com/w/cpp/container/list/splice). See https://www.youtube.com/watch?v=W2tWOdzgXHA&t=45m56s. What do you think?

PR #249, opened 2021-05-08
p7zip patches

F-TD5X:
add patches till CVE-2021-3465
add patch for build on linux


F-TD5X:
These patches only tested on `archlinux:latest` and `alpine:latest` (pulled at 2021-04-27).

luisangelsm:
This breaks macos builds

F-TD5X:
Sorry for my wrong use of `@@`, I have update it for patch command to work.

selmf:
No worries, we all make mistakes. That is one of the reasons we have CI.

@luisangelsm build works fine now, could you run a quick check if file support works as intended on macos?

luisangelsm:
It seems to work.

PR #263, opened 2021-06-20
Clazy autorefactoring for SIGNAL/SLOT to new syntax

selmf:
This is the first batch of a rather huge PR with the intention of getting rid of the old Qt SIGNAL/SLOT syntax. Most of the heavy lifting was done using the Clazy static code checker for Qt code, with some minor manual corrections to the code.

The suggested way to introduce these changes is to cherry-pick the **first three commits** which are mostly automated, safe replacements for the old connections and to introduce them ASAP to reduce the risk of this work getting outdated and to avoid risky labor intensive rebase work.

As the conversions after the first batch of automated fixes are mostly manual, they have a higher risk to introduce bugs and thus should be independently verified and tested before being introduced into the code base.

**This PR is not intended to be merged as a whole, but to serve as a testing and staging area for this work.**

I will introduce additional commits to this PR as I continue the conversion to the new syntax.

### DIFF
--- a/YACReader/bookmarks_dialog.cpp
+++ b/YACReader/bookmarks_dialog.cpp
@@ -69,7 +69,7 @@ BookmarksDialog::BookmarksDialog(QWidget *parent)
 
     cancel = new QPushButton(tr("Close"));
     cancel->setFlat(true);
-    connect(cancel, SIGNAL(clicked()), this, SLOT(hide()));
+    connect(cancel, &QAbstractButton::clicked, this, &QWidget::hide);
     buttons->addStretch();
     buttons->addWidget(cancel);
 
--- a/YACReader/goto_dialog.cpp
+++ b/YACReader/goto_dialog.cpp
@@ -21,9 +21,9 @@ void GoToDialog::setupUI()
     textLabel->setAlignment(Qt::AlignRight | Qt::AlignVCenter);
 
     accept = new QPushButton(tr("Go To"));
-    connect(accept, SIGNAL(clicked()), this, SLOT(goTo()));
+    connect(accept, &QAbstractButton::clicked, this, &GoToDialog::goTo);
     cancel = new QPushButton(tr("Cancel"));
-    connect(cancel, SIGNAL(clicked()), this, SLOT(close()));
+    connect(cancel, &QAbstractButton::clicked, this, &QWidget::close);
 
     auto topLayout = new QHBoxLayout;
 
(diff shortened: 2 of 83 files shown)


selmf:
Alright, time to tackle this. This PR consists of two parts - automatic or mostly automatic conversions and manual rewrites. The automatic conversions are in the first three commits and should be safe, so there is no need to test every signal conversion. These are the commits up to and including  https://github.com/YACReader/yacreader/pull/263/commits/4539b2a99dc620fc81198cd4524ab63830536d31 .

What we need to test thoroughly are:

- connections containing QOverload
- connections containing a lambda expression
- anything that is only relative to a single OS

In quite a few cases, manual work was also needed because the base classes default signals are overwritten by our derived classes. While this is technically possible, it is generally considered bad practice and can lead to undefined behavior and bugs that are hard to catch. We will have to tackle that in another pull request.

Currently, only https://github.com/YACReader/yacreader/pull/263/commits/009d8cf364bb5293babaee7070ee0cc585533f14 is tested and confirmed (thanks @luisangelsm )

This table will be used to track the status of verification:

|Commit|Test status|Result|
|---------|-----------|-------|
|https://github.com/YACReader/yacreader/pull/263/commits/a71aec1cee3b1408d55c65afb3a390a4ec6fa4b4| - | auto |
|https://github.com/YACReader/yacreader/pull/263/commits/3f1075ce405c845fffc40ddfcfac42cef175db49| - | auto |
|https://github.com/YACReader/yacreader/pull/263/commits/be7f65a88fe04444151718e7fe08833be13977ac| - | auto |
|https://github.com/YACReader/yacreader/pull/263/commits/be7f65a88fe04444151718e7fe08833be13977ac| in Progress| ? |
|https://github.com/YACReader/yacreader/pull/263/commits/009d8cf364bb5293babaee7070ee0cc585533f14| done | works |
|https://github.com/YACReader/yacreader/pull/263/commits/f7c64a3a18c3af850f83a2786bbea93343f91248| | |
|https://github.com/YACReader/yacreader/pull/263/commits/d976e0cc84e196ebe0c29cbdc8d4f0de17dde37d| | |
|https://github.com/YACReader/yacreader/pull/263/commits/c4cfd5b5d4c19baf3dfc4049c213aa2987ec1a52| | |
|https://github.com/YACReader/yacreader/pull/263/commits/669b002f0c7598dd8631cbf5c24f4c0c54012fe5| | |
|https://github.com/YACReader/yacreader/pull/263/commits/b42425d64cb7a5eaf969b8f15f454e45a879861c| | |
|https://github.com/YACReader/yacreader/pull/263/commits/4d6e32e6fd9105578f47e8faab88a53c91aae8dd| | |
|https://github.com/YACReader/yacreader/pull/263/commits/1ae4c1e68e5058d51272b96f74d4b5252d8d1bbc| | |
|https://github.com/YACReader/yacreader/pull/263/commits/afc931ecd963f3d041c8578a6b14f6f2e724d5a1| | |

Issue #31, opened 2018-09-12
version numbers missing on macOS


core-code:
i've downloaded Version '9.5.0' but when selecting it in the Finder and using 'Get Info' it doesn't show any Version-number at all. i guess the CFBundleVersion and CFShortVersionString entries in your Info.plist file are missing

selmf:
Is this still an issue with 9.6.2?

core-code:
yes. both YACReader and YACReaderLibrary don't have any version numbers.

the CFBundleVersion and CFShortVersionString entries in your Info.plist file are missing

selmf:
I'm not really a OS X guy but sorting out version setting for our build system and CI is currently on my agenda, so I'll look into it. Not sure if there is a straightforward way to do his with our build system though.
Do you happen to have some documentary for Info.plist and the required entries handy?

core-code:
there is some documentation here:
https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html

however, i'd recommend comparing the Info.plist with something that Xcode generates for a sample app, as this is like the "best practices".

i think the most important thing is filling out the version numbers, as mentioned, and also importantly removing the 'CFBundleGetInfoString' entry. this one has been deprecated for more than 10 years, and will actually prevent display of the version number in the Finder, even if you do add it.

luisangelsm:
@selmf I will send you the keys you need to set in the Info.plist. I totally forgot about this while I was setting up the azure pipelines.

selmf:
Thanks. If possible I'd like Info.plist to be generated from a template which is configured by our build system instead of having hardcoded version entries.

vitorgalvao:
To add to this, technically you don’t *have to* add the version during the build process, you can add it after it’s done by directly editing the `Info.plist`:

```bash
# Change these values
app_path='/PATH/TO/APP/HERE'
app_version='VERSION NUMBER HERE'

# Keep these as they are
/usr/libexec/PlistBuddy -c "Add :CFBundleVersion string ${app_version}" "${app_path}/Contents/Info.plist"
/usr/libexec/PlistBuddy -c "Add :CFBundleShortVersionString string ${app_version}" "${app_path}/Contents/Info.plist"
```

[...]

PR #264, opened 2021-06-25
compileOSX.sh: add version to app

vitorgalvao:
Followup to https://github.com/YACReader/yacreader/issues/31#issuecomment-866274780. Untested in your setup, but it should work if I’m understanding it correctly.

Fixes https://github.com/YACReader/yacreader/issues/31.

### DIFF
--- a/YACReader/Info.plist
+++ b/YACReader/Info.plist
@@ -8,8 +8,6 @@
 	<string>YACReader.icns</string>
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
-	<key>CFBundleGetInfoString</key>
-	<string>YACReader</string>
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleExecutable</key>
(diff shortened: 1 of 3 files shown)


luisangelsm:
Thanks for the PR, the version and build number are now in the info.plist file, but "Get Info" doesn't show them

image

vitorgalvao:
> the version and build number are now in the info.plist file, but "Get Info" doesn't show them

Should be fixed by the commit I just pushed, removing [the obsolete CFBundleGetInfoString](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html). As [explained by @core-code](https://github.com/YACReader/yacreader/issues/31#issuecomment-546636492):

> importantly removing the 'CFBundleGetInfoString' entry. this one has been deprecated for more than 10 years, and will actually prevent display of the version number in the Finder, even if you do add it.

vitorgalvao:
I see one of the checks is failing, but it’s on Windows so I’m assuming it’s unrelated to this PR or was a glitch.

vitorgalvao:
All checks passing now.

luisangelsm:
Yeah, everything is working now, thanks for PR!!

PR #279, opened 2021-10-19
Mingw compatibility

joankaradimov:
These are bunch of small tweaks that allow the building of YACReader under MinGW on Windows. I tested this under MSYS2.


joankaradimov:
@selmf Since you're considering an msys2 build, I'm adding [another commit](https://github.com/YACReader/yacreader/pull/279/commits/d168a2975e266b9a43e43bfd285bb76f8b6dcb3d) that alllows the usage of `pkg-config` in mingw builds.

luisangelsm:
@joankaradimov  thanks for your work!

@selmf I think that adding a build for mingw would keep things sane in the long term, otherwise someone will break it sooner or later. The only thing that worries me is that our windows builds in CI cause false positives because problems installing dependencies, so having 3 of them will increase the chances of a failure in CI.

PR #283, opened 2021-11-09
Add updated mime-types to YACReader.desktop

bsdf:
according to the IANA, [`application/vnd.comicbook+zip`](https://www.iana.org/assignments/media-types/application/vnd.comicbook+zip) and [`application/vnd.comicbook-rar`](https://www.iana.org/assignments/media-types/application/vnd.comicbook-rar) are the correct mimetypes for cbz and cbr. i was having issues getting `xdg-open` to open my cbz and this fixed the issue


selmf:
Good catch, I didn't know these mime types existed. Could you add an entry in our changelog to round it up?

bsdf:
updated

selmf:
Merged, thank you

Issue #184, opened 2021-01-01
Option to view the whole library as a homepage


Warden20:
It would be awesome if there is a option to set the homepage. I have my comics in folders but is annoying to open one folder at a time

PR #285, opened 2021-11-09
Add ability to view whole library

noku:
Implements: https://github.com/YACReader/yacreader/issues/184

### Summary

- Add an additional option `All` to the `TypeSpecialList`. The purpose of the option is to show all the comics.
- This option is helpful in providing an overview of all the libraries and making it easier to access comics without opening each folder one by one. This should prove extremely helpful for big libraries.

|          Preview           |
| :------------------------: |
|  |

### Alternative Approach

Show the library content (all comics) on the main view. (instead of subfolder information). 
  - This approach will have a bigger impact on the user experience. That's why I opted for a more simple approach. (unless you find this alternative approach acceptable)

### Additional Notes

- Currently, this PR has the additional purpose of building and testing the approach using Azure Pipeline since the attempt to set up the project locally failed. 
-  Non-related: For some reason, the scroll functionality in `YacreaderLibrary` `dmg` files generated by Azure does not work.


noku:
Closed in favor of https://github.com/YACReader/yacreader/pull/285

PR #286, opened 2021-11-10
Add ability to view entire library

noku:
Implements: https://github.com/YACReader/yacreader/issues/184

### Summary

- Add the ability to view the entire library. Add an additional item in the folders menu bar. 
- This option is helpful in providing an overview of all the folders/comics and making it easier to access them without opening each folder one by one. This should prove extremely helpful for big libraries.

##
- Unfortunately for some unknown reason, the icon image doesn't show in the application built through CI.

### Additional Notes

- Currently, this PR has the additional purpose of building and testing the approach using Azure Pipeline since the attempt to set up the project locally failed.


noku:
No longer necessary, found an workaround by typing `:` in the search bar, should cover the needs for now

PR #300, opened 2022-04-21
Add AppImage generator for YACReader and YACReaderLibrary

popsUlfr:
Here's a generator script that can create AppImages of YACReader. It bundles `pdfium` and the latest `unarr` with 7zip support.

This could be plugged into CI to produce AppImages for new releases.

### DIFF
--- a/appimage/README.md
+++ b/appimage/README.md
@@ -1,0 +1,14 @@
+# AppImage
+
+To build the AppImages of `YACReader` and `YACReaderLibrary`, podman or docker is required using an ubuntu base image of 20.04:
+
+```
+podman pull ubuntu:20.04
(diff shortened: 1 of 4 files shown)


luisangelsm:
Anyone interested in adding a CI job that uses this? I would be nice to have it ready for the next release.

luisangelsm:
Not sure how hard would be, but I would prefer to have p7zip instead of unarr as the decompression backend, mainly because of rar5 support.

MetaStazz:
As I love AppImage, I tried to make one for YAC, as I build it with libarchive and poppler.
But yeah, is not that easy to build one with multiple binaries !!

PR #335, opened 2022-10-13
Create Flatten_a_Linked_List.cpp

Mahesmati:
i solved this problem using cpp. kindly consider it. It's my first pull resuest


luisangelsm:
PRs need to be related to the project.

PR #338, opened 2022-10-14
Added unfinished German translations

CandidDeer:
As requested, I have added all unfinished German translations.

### DIFF
--- a/YACReaderLibrary/yacreaderlibrary_de.ts
+++ b/YACReaderLibrary/yacreaderlibrary_de.ts
@@ -5,7 +5,7 @@
     <name>ActionsShortcutsModel</name>
     <message>
         <source>None</source>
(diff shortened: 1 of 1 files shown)


selmf:
I did a quick check and it looks good. Wording is much more natural and it reads like an actual translation now. Great job!

I spotted a small typo (YACREader instead of YACReader) but otherwise did not a deeper analysis.

@luisangelsm Once the typo is fixed we should merge this.

CandidDeer:
@selmf Thank you for the heads-up on the typo. I've fixed it now. Also, thank you for liking my translation. As good as computer translation has become in the past few years, I find that (for now) nothing beats a real human translation. ✨

selmf:
Merged. Thank you for the contribution :)

Yes, translation is a sometimes underappreciated art.

PR #342, opened 2022-10-23
Added translations for some missing strings

fadkeabhi:
Added translation for some missing strings


selmf:
@luisangelsm don't merge this, it is garbage machine translation.

PR #355, opened 2022-11-12
Fix crash when exiting YACReader while processing a comic

selmf:


selmf:
This is a fix for crashes that can happen when YACReader is closed while still processing a comic. When the comic thread is not properly finalized and gets deleted before finishing, segfaults and worse problems can occur.

As some of the errors fixed in this PR were masked by the logger not being able to receive messages when the main application thread has finished, logging is moved to a dedicated thread.

PR #356, opened 2022-11-13
Add support for poppler-qt6 pdf backend

selmf:
This PR adds support for the Poppler Qt6 backend

@jbeich feel free to give it a test, feedback is always welcome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Version counting is based on semantic versioning (Major.Feature.Patch)
 
 ### All apps
 * Run logger in a dedicated thread to avoid segfaults at application shutdown
+* Add support for poppler-qt6 pdf backend
 
 ## 9.10
 
--- a/YACReaderLibrary/initial_comic_info_extractor.cpp
+++ b/YACReaderLibrary/initial_comic_info_extractor.cpp
@@ -27,7 +27,7 @@ void InitialComicInfoExtractor::extract()
 #ifndef NO_PDF
     if (fi.suffix().compare("pdf", Qt::CaseInsensitive) == 0) {
 #if defined Q_OS_MAC && defined USE_PDFKIT
-       auto pdfComic = std::make_unique<MacOSXPDFComic>();
+        auto pdfComic = std::make_unique<MacOSXPDFComic>();
         if (!pdfComic->openComic(_fileSource)) {
             return;
         }
@@ -64,7 +64,6 @@ void InitialComicInfoExtractor::extract()
                 saveCover(_target, p);
             } else if (_target != "") {
                 QLOG_WARN() << "Extracting cover: requested cover index greater than numPages " << _fileSource;
-
             }
         }
         return;
--- a/YACReaderLibrary/initial_comic_info_extractor.cpp
+++ b/YACReaderLibrary/initial_comic_info_extractor.cpp
@@ -37,8 +37,12 @@ void InitialComicInfoExtractor::extract()
             return;
         }
 #else
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+        auto pdfComic = Poppler::Document::load(_fileSource);
+#else
         auto _pdfComic = Poppler::Document::load(_fileSource);
-        auto pdfComic = std::unique_ptr<Poppler::Document>(_pdfComic);
+        pdfComic = std::unique_ptr<Poppler::Document>(_pdfComic);
+#endif
 #endif
         if (!pdfComic) {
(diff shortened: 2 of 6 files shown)


jbeich:
Fails to build on FreeBSD 13.1. Clang is default on FreeBSD, OpenBSD, macOS, Android, OpenMandriva, ChimeraLinux. libc++ is often used together with Clang and generally cannot be mixed with libstdc++ due to conflict between libc++abi (or libcxxrt) and libsupc++. Given YACReader depends on C++-based libraries (Qt6, Poppler) it cannot be built with different C++ library than the default (unless those are bundled).

Clang/libc++
```c++
YACReaderLibrary/initial_comic_info_extractor.cpp:41:25: error: functional-style cast from 'std::unique_ptr<Poppler::Document>' to 'std::unique_ptr<Poppler::Document>' uses deleted function
        auto pdfComic = std::unique_ptr<Poppler::Document>(_pdfComic);
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/v1/__memory/unique_ptr.h:121:59: note: candidate constructor (the implicit copy constructor) has been implicitly deleted
class _LIBCPP_UNIQUE_PTR_TRIVIAL_ABI _LIBCPP_TEMPLATE_VIS unique_ptr {
                                                          ^
```
GCC/libc++
```c++
YACReaderLibrary/initial_comic_info_extractor.cpp: In member function 'void YACReader::InitialComicInfoExtractor::extract()':
YACReaderLibrary/initial_comic_info_extractor.cpp:41:69: error: use of deleted function 'constexpr std::__1::unique_ptr<Poppler::Document>::unique_ptr(const std::__1::unique_ptr<Poppler::Document>&)'
   41 |         auto pdfComic = std::unique_ptr<Poppler::Document>(_pdfComic);
      |                                                                     ^
In file included from /usr/include/c++/v1/__memory/shared_ptr.h:25,
                 from /usr/include/c++/v1/__functional/function.h:20,
                 from /usr/include/c++/v1/functional:500,
                 from /usr/include/c++/v1/optional:154,
                 from /usr/local/include/qt6/QtCore/qtypeinfo.h:8,
                 from /usr/local/include/qt6/QtCore/qglobal.h:1397,
                 from /usr/local/include/qt6/QtCore/QtCore:4,
                 from /usr/local/include/qt6/QtGui/QtGuiDepends:3,
                 from /usr/local/include/qt6/QtGui/QtGui:3,
                 from ../YACReaderLibrary/initial_comic_info_extractor.h:4,
                 from ../YACReaderLibrary/initial_comic_info_extractor.cpp:1:
/usr/include/c++/v1/__memory/unique_ptr.h:121:59: note: 'constexpr std::__1::unique_ptr<Poppler::Document>::unique_ptr(const std::__1::unique_ptr<Poppler::Document>&)' is implicitly declared as deleted because 'std::__1::unique_ptr<Poppler::Document>' declares a move constructor or move assignment operator
  121 | class _LIBCPP_UNIQUE_PTR_TRIVIAL_ABI _LIBCPP_TEMPLATE_VIS unique_ptr {
      |                                                           ^~~~~~~~~~
```

selmf:
Well I am glad I asked you to test this now. 

What is happening here is not a compiler or libstdc++ problem but a mistake on my side.

The compile error is caused by the compiler hitting a line it shouldn't. What is missing here is the #ifdef guard and the poppler-qt6 specific logic for the unique pointer, which is different from poppler-qt5.

selmf:
Ok, I pushed a fix for the problem and I did a test build with clang, qt6 and poppler-qt6 on my Linux machine.

jbeich:
725cf658 builds fine with both [poppler-qt6](https://github.com/YACReader/yacreader/files/10042130/yacreader-qt6-9.10.0.13.log) and [poppler-qt5](https://github.com/YACReader/yacreader/files/10042138/yacreader-9.10.0.13.log). Tested runtime via a PDF version of FreeBSD Journal: dssim for poppler-qt6 (fad24a5) vs. poppler-qt5 (9.10.0) returns 0.00000000 (identical) on a screenshot of a single page.

luisangelsm:
@selmf maybe we should add a new job(s) in Azure Pipelines for poppler-qt6 to catch problems like the one @jbeich found (thanks for testing btw).

luisangelsm:
@selmf could you rebase this on top of develop please?

selmf:
@luisangelsm rebase is up.

Adding a test build for poppler-qt6 to catch errors is absolutely something we need to do. The problem here is that Ubuntu does not package poppler-qt6 yet, so I could not add a test to our build pipelines.

PR #403, opened 2023-10-22
ComicInfo::getTypeString: Return an empty string for unhandled enums

selmf:
When creating switch case structures, please don't forget the default case or you can have invalid control flows in your code :)

### DIFF
--- a/YACReaderLibrary/db/comic_model.cpp
+++ b/YACReaderLibrary/db/comic_model.cpp
@@ -1307,6 +1307,9 @@ void ComicModel::deleteComicsFromSpecialList(const QList<QModelIndex> &comicsLis
     case ReadingListModel::TypeSpecialList::Favorites:
         deleteComicsFromFavorites(comicsList);
         break;
+    case ReadingListModel::TypeSpecialList::Recent:
+        // do nothing, recent is read only
+        break;
     }
 }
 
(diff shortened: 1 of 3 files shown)


luisangelsm (changes requested):
`FileType` is an enum class, so the compiler will take care of missing branches if `FileType` ever changes (although our current compiler flags only raises a warning, this should be an error), adding default won't let us catch that this switch needs to handle any new values. If you keep adding defaults to all the switches for enum classes, updating them can become a major headache in the future. 

A more appropriate fix for this method would be to use `canConvert` on `type` to see if the underlying value is messed up.

selmf:
I've adjusted the code so the enum warnings are not suppressed and added a return QString() to stop the compiler from complaining about control flow issues in the code.

For good measure, I also converted the chars to QStringLiteral. For more info, see here: https://doc.qt.io/qt-6/qstring.html#QStringLiteral

selmf:
I've added compiler flags to promote the warnings to errors for unhandled enums in switch case statements.

Please note that this will probably not  build as there is at least one such incomplete switch present in our code :man_shrugging:

luisangelsm:
That compiler flags goes a little too far :/

image

luisangelsm:
> For good measure, I also converted the chars to QStringLiteral. For more info, see here: https://doc.qt.io/qt-6/qstring.html#QStringLiteral

I didn't see those changes in the code

selmf:
It seems like I made an error when doing the commit. It was supposed to look like this:

`QString ComicInfo::getTypeString()
{
    switch (type.value<YACReader::FileType>()) {
    case YACReader::FileType::Comic:
        return QStringLiteral("Comic");
    case YACReader::FileType::Manga:
        return QStringLiteral("Manga");
    case YACReader::FileType::WesternManga:
        return QStringLiteral("Western Manga");
    case YACReader::FileType::WebComic:
        return QStringLiteral("Web Comic");
    case YACReader::FileType::Yonkoma:
        return QStringLiteral("4-Koma");
    }
    return QString();
}`

PR #428, opened 2024-03-02
Compile in linux using poppler

luisangelsm:
Having the option there can help packagers to do the right thing and build YACReader with PDF support.


luisangelsm:
Done in the sample linux build scripts

PR #459, opened 2024-12-14
Bump Qt to 6.9.x in Windows and macos

luisangelsm:


luisangelsm:
Done as part of the 9.16 release

PR #499, opened 2026-01-11
Add Qt image plugins for AVIF, JXL, and HEIC/HEIF support (Docker)

nickv2002:
# Add Qt Image Plugins for Modern Image Format Support (Docker)

## Summary

This PR adds support for modern image formats (AVIF, JXL, HEIC/HEIF) to the YACReaderLibraryServer Docker image using Qt image plugins. **No source code changes are required** - this follows YACReader's existing plugin architecture as suggested in #498.

## Motivation

Modern image formats like AVIF and JPEG XL offer significant advantages:
- **Better compression**: 30-50% smaller file sizes vs JPEG at same quality
- **Higher quality**: Better detail preservation at lower bitrates  
- **Modern features**: HDR, lossless compression, progressive decoding
- **Growing adoption**: Increasingly used in digital comics and manga

Users with comics containing these formats currently cannot view them in YACReader, despite the formats becoming more common.

## Implementation

This PR follows the maintainer's guidance from #498 to use Qt's dynamic plugin loading system rather than adding custom decoders to the codebase.

### Docker Multi-Stage Build (8 Stages)

The updated Dockerfile uses an optimized multi-stage architecture with a shared base layer for maximum build caching and parallelization:

**Stage 0-1: Base Layers**
- **base** - Common Ubuntu noble base image
- **plugin-base** - Shared build dependencies (cmake, git, qt6-base-dev, etc.) inherited by all plugin builders

**Stage 2-5: Independent Plugin Builders** (all FROM plugin-base, build in parallel)
- **sevenzip-builder** - Builds 7z.so with RAR support
- **heic-plugin-builder** - Builds qt-heic-image-plugin
- **jxl-plugin-builder** - Builds libjxl 0.10.2 + qt-jpegxl-image-plugin (merged into one stage)
- **avif-plugin-builder** - Builds qt-avif-image-plugin

**Stage 6-7: Assembly**
- **yacreader-builder** - Compiles YACReaderLibraryServer, collects plugins from stages 2-5
- **runtime** - Final minimal image with only runtime dependencies

**Key Optimizations:**
- **Shared plugin-base stage**: Common dependencies installed once, reused by all builders
- **Maximum parallelization**: Stages 2, 3, 4, 5 build simultaneously
- **Merged libjxl stages**: libjxl build + JXL plugin build in single stage (no intermediate copy)
- **Better caching**: Changes to one plugin don't invalidate others
- **Smaller layers**: No duplicate dependency installations

### Qt Image Plugins Used

[...]

PR #505, opened 2026-02-17
feat(docker): Add KDE kimageformats for AVIF, JXL and additional image formats

nickv2002:
## Description

Adds KDE kimageformats support to the Docker image by rebasing on Ubuntu 26.04 LTS (Resolute), which ships KF6 natively. This enables AVIF, JXL, HEIC, and 20+ additional image formats in YACReaderLibraryServer with no source builds required.

### The key change

Ubuntu Noble (24.04) only ships KF5 — getting KF6 kimageformats required building ECM and kimageformats from source, adding significant build complexity. Ubuntu Resolute (26.04 LTS) ships `kimageformat6-plugins` 6.24.0 directly in apt, so it installs as a single runtime dependency alongside the existing packages.

### Changes

- Rebase on `ghcr.io/linuxserver/baseimage-ubuntu:resolute` (Ubuntu 26.04 LTS, Qt 6.10)
- Add `kimageformat6-plugins` to the runtime apt install — that's it
- Reorganize Dockerfile into a cleaner multi-stage build: `sevenzip-builder` → `yacreader-builder` → `runtime`
- Fix `COPY docker/root.tar.gz /` path (build context is the repo root)

### Supported Formats Added

✅ **AVIF / AVIFS** — via libavif  
✅ **JXL** — via libjxl  
✅ **HEIC / HEIF** — via libheif (included in kimageformat6-plugins on Resolute)  
✅ **Bonus formats** — ani, bw, eps, hdr, ico, pcx, pic, psd, qoi, ras, rgb, tga, xcf, kra, ora, and more

### Testing

- ✅ Docker image builds successfully on Ubuntu 26.04 / Qt 6.10
- ✅ AVIF and JXL formats recognized by YACReaderLibraryServer at startup
- ✅ Test comics scanned successfully: avif.cbz, jxl.cbz, test_combined.cbz
- ✅ RAR (CBR) support verified working
- ✅ Production tested for several weeks

### Notes

- **Dockerfile.aarch64** is not modified in this PR — ARM64 support can be addressed as a follow-up
- Compared to PR #499, this approach uses the official KDE kimageformats package from apt: no source builds, smaller diff, more formats, easier to maintain

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -132,7 +132,7 @@ ENV LC_ALL="en_US.UTF-8" \
     PATH="/app/bin:${PATH}"
 
 # copy files
-COPY docker/root.tar.gz /
+COPY root.tar.gz /
 # Extract the contents of root.tar.gz into the / directory
 RUN tar -xvpzf /root.tar.gz -C /
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,34 @@
-FROM ghcr.io/linuxserver/baseimage-ubuntu:noble AS builder
+FROM ghcr.io/linuxserver/baseimage-ubuntu:resolute AS base
+
+# ============================================================================
+# Stage: sevenzip-builder
+# ============================================================================
+FROM base AS sevenzip-builder
+
+# get 7zip source
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+       7zip \
+       build-essential \
+       make \
+       wget && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN cd /tmp && \
+    wget "https://github.com/YACReader/yacreader-7z-deps/blob/main/7z2301-src.7z?raw=true" -O 7z2301-src.7z && \
+    7z x 7z2301-src.7z -o/tmp/lib7zip
+
+# install 7z.so with RAR support
+RUN cd "/tmp/lib7zip/CPP/7zip/Bundles/Format7zF" && \
+    make -f makefile.gcc && \
+    mkdir -p /app/lib/7zip && \
+    cp ./_o/7z.so /app/lib/7zip
+
+# ============================================================================
+# Stage: yacreader-builder
+# ============================================================================
+FROM base AS yacreader-builder
 
 # repository URL and version, which can be a tag, branch, or commit SHA
 ARG YACR_REPOSITORY="https://github.com/YACReader/yacreader.git"
@@ -28,8 +58,6 @@ RUN \
        qt6-tools-dev-tools \
        libgl-dev \
        qt6-l10n-tools \
-       libqt6opengl6-dev \
-       libunarr-dev \
        qt6-declarative-dev \
        libqt6svg6-dev \
        libqt6core5compat6-dev \
@@ -55,7 +83,6 @@ RUN \
        zlib1g-dev && \
     ldconfig
 
-
 # clone YACReader repo
 RUN git clone "$YACR_REPOSITORY" /src/git && \
     cd /src/git && \
@@ -72,15 +99,13 @@ RUN cd /src/git && \
     cmake --build build --parallel && \
     cmake --install build
 
-# install 7z.so with RAR support
-RUN echo "Building and installing 7z.so with RAR support..." && \
-    cd "/src/git/compressed_archive/lib7zip/CPP/7zip/Bundles/Format7zF" && \
-    make -f makefile.gcc && \
-    mkdir -p /app/lib/7zip && \
-    cp ./_o/7z.so /app/lib/7zip
+# Use pre-built 7z.so from sevenzip-builder (provides RAR support)
+COPY --from=sevenzip-builder /app/lib/7zip /app/lib/7zip
 
-# Stage 2: Runtime stage
-FROM ghcr.io/linuxserver/baseimage-ubuntu:noble
+# ============================================================================
+# Stage: runtime
+# ============================================================================
+FROM base AS runtime
 
 # env variables
 ENV APPNAME="YACReaderLibraryServer"
@@ -88,7 +113,7 @@ ENV HOME="/config"
 LABEL maintainer="luisangelsm"
 
 # Copy the built application from the builder stage
-COPY --from=builder /app /app
+COPY --from=yacreader-builder /app /app
 
 # runtime packages
 RUN apt-get update && \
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -1,4 +1,90 @@
...
+FROM plugin-base AS kimageformats-builder
+
+# Install additional dependencies for kimageformats
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    libavif-dev \
+    libjxl-dev \
+    libarchive-dev && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
```
nickv2002: redundant to have multiple apt-get phases here even because plugin base is only used in 1 place, just get rid of plugin-base step and merge it into kimageformats-builder

luisangelsm:
@nickv2002 thanks! did you try to install kimageformat-plugins instead of building them? like adding `kimageformat-plugins` to the list of packages installed.

nickv2002:
> @nickv2002 thanks! did you try to install kimageformat-plugins instead of building them? like adding `kimageformat-plugins` to the list of packages installed.

I tried the versions from `apt` but:

> Build from source required: Both ECM 6.0.0 and kimageformats 6.0.0 must be built from KDE source because Ubuntu Noble only provides older KF5 framework versions (ECM 5.115.0). KF6 framework is required for AVIF/JXL support.

Could rebase the image on a different & newer distro but that would introduce other changes.

luisangelsm:
> Could rebase the image on a different & newer distro but that would introduce other changes.

What changes? Any downsides?

nickv2002:
> > Could rebase the image on a different & newer distro but that would introduce other changes.
> 
> What changes? Any downsides?

That would require significantly reworking because the base docker image to a more leading edge version of alpine/fedora/arch. It also doesn't seem in keeping with your previously expressed interest in minimizing maintenance work and support burdens.

selmf:
I think we could use a newer non-LTS Ubuntu base image.

There's also another option I have used before - the KDE Neon Distro is based on Ubuntu LTS (noble) and should have the latest Qt and kimageformats. We could add the apt/ppas and just use those.

nickv2002:
Ubuntu 26.04 is an LTS release and should be out in about 2 months. That should have built-in ECM 6.* and kimageformats 6.* in the standard apt library. Less work than some alternate repo and includes long term support.
I'll check back again then in a few months once that's out.

nickv2002:
## 7zip Build Optimization Update

The latest commit adds an important optimization to the Docker build process by simplifying the 7zip plugin handling:

### What Changed
- **Removed:** Custom source build of `7z.so` (was using `make -f makefile.gcc` on 7zip SDK source)
- **Replaced with:** Pre-built apt packages (`7zip` + `7zip-rar`) from Ubuntu repositories

### Key Details

**Files being used:**
- `7zip` package → `/usr/lib/7zip/7z.so` (2.6MB main format handler)
- `7zip-rar` package → `/usr/lib/7zip/Codecs/Rar.so` (111KB RAR5 codec)

**Why we still download the source:**
The 7zip SDK source archive (`7z2301-src.7z`) is still downloaded and extracted because YACReader's lib7zip C++ wrapper needs the SDK headers at compile-time. Ubuntu's 7zip apt package doesn't include development headers, only binaries.

**Build process now:**
1. Download 7zip SDK source (for headers)
2. **Extract** it (unchanged)
3. **Skip** the compilation step (time saved: 15-20 min → 5-10 min)
4. Copy pre-built `.so` files from apt instead

### Benefits
✅ Simplifies the build process (less YACReader-specific code to maintain)  
✅ Faster builds (no compilation of 7z.so required)  
✅ Better RAR support (via unrar library in apt package)  
✅ Reduces dependencies (removes build-essential/make from sevenzip-builder)  
✅ Production tested for 1 week without issues

### Testing Status
- ✅ Container rebuilt and running in production
- ✅ YACReaderLibraryServer confirms: "Compression backend: 7zip"
- ✅ Both `7z.so` and `Codecs/Rar.so` present and accessible

docker/Dockerfile
```diff
@@ -71,32 +98,31 @@ RUN cd /src/git && \
...
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
        libqt6core5compat6 \
        libpoppler-qt6-3t64 \
        qt6-image-formats-plugins \
-       libqt6network6t64 \
-       libqt6sql6-sqlite && \
+       libqt6network6 \
+       libqt6sql6-sqlite \
+       kimageformat6-plugins && \
```
nickv2002, line 126: Yay, easy to add lots of new formats!

docker/Dockerfile
```diff
@@ -105,7 +131,7 @@ ENV LC_ALL="en_US.UTF-8" \
     PATH="/app/bin:${PATH}"
 
 # copy files
-COPY root.tar.gz /
+COPY docker/root.tar.gz /
```
nickv2002: I think this is actually a bug fix, but it all depends on where your relative path is anyway, so I can revert it if it messes up the build scripts online.

nickv2002:
@luisangelsm @selmf - check out these changes. Now building off of Ubuntu Resolute instead of Noble (that's 26.04 instead of 24.04) so we can pull in the latest image formats as a plug-in from `apt` without needing to build anything. Most of the diff here is just to put the 7Zip build in its own build step for parallelism + build speed.

LMK and can port these changes over to the ARM64 image too.

docker/Dockerfile
luisangelsm: #14 [runtime 3/4] COPY docker/root.tar.gz /
#14 ERROR: failed to calculate checksum of ref e27c7740-8587-4822-9bec-fa47c65630a3::7jnksf1kizgh6nx3atdvq7ih0: "/docker/root.tar.gz": not found

luisangelsm:
Thanks for the update @nickv2002 

CI is failing. I think it's the `docker/root.tar.gz` change.

docker/Dockerfile
nickv2002: Thanks — fixed in 58e6ae6b. Reverted to `COPY root.tar.gz /`

luisangelsm:
@nickv2002  CI passes now, so you can update the arm64 Dockerfile.

nickv2002:
@luisangelsm I updated this PR include the same updates for the arm64 Dockerfile as the x86_64 file. I built the ARM version on my M-series Mac and used it to serve some of the same JXL/AVIF/etc sample comics I used for x86_64 testing. Worked fine.

----------

Follow-up: opened #540 to consolidate `Dockerfile.aarch64` into the shared `Dockerfile`.

After the last commit here (`refactor(docker): Split 7zip build into its own stage for aarch64`) the two Dockerfiles are byte-identical — same `resolute` base (a multi-arch manifest), same package lists, same multi-stage layout. #540 deletes the duplicate and points the existing `docker-arm64` CI job at the shared file with `--platform linux/arm64` (mirroring the explicit `--platform linux/amd64` already on the amd64 job). The two-job-plus-manifest-combine CI structure is unchanged, so multi-arch publishing keeps working.

Kept separate from this PR because the scope is different: this PR is "make arm64 work on resolute + AVIF/JXL" (functional), #540 is "remove the now-redundant duplicate file" (cleanup). #540 is stacked on this branch and needs to merge after this one.

luisangelsm:
Thanks @nickv2002 :)

PR #514, opened 2026-02-24
Add "Delete current comic and open next" shortcut (Ctrl+Del)

dragonpaw:
Adds a new Ctrl+Delete shortcut that moves the currently open comic to the OS recycle bin (with permanent deletion as fallback) and automatically opens the next comic in sequence. Falls back to previous comic if no next exists, or resets the viewer if no siblings remain.

When reading new comics, I tend to know within a few pages if I want to keep it or not. This allows me to quickly jump to the next comic, and is likely to be quite safe as it's a hard keystroke to accidentally hit.


luisangelsm:
This needs to target the YACReader 10 branch, a lot it's changing and `yacreader10` is basically the new develop until it gets refined enough to get merged into `develop`.

I need to think about this though, I am not a big fan of adding destructive actions to YACReader. Ideally this should notify YACReaderLibrary if the file was opened from it, so YACReaderLibrary takes care of deleting the file and it can update the db.

dragonpaw:
I can definitely redo it for v10 and add the notification to Library. 

I understand the caution around potentially destructive actions. I tried to make it safe by using a really unlikely keystroke and by routing comics to the Recycle Bin so you can get them back if you change your mind. One potential even further step would be not having a default keybind for the action. Which would require an action by the user to opt-in by adding a keybind. Or maybe a popup when it's hit with "Do you want to delete this comic?" and "Yes", "Yes, Don't ask again" and "No", as choices.

I'm a bit of a digital pack-rat, and tend to assume online content can vanish tomorrow if I don't save it. So I will end up with hundreds or more comic files and then have to keep the 5% of them I actually like. Right now I have... 8.954 to go through, of which I might keep 200 tops. (Don't judge me. 🥲 ) 

Part of the notion of open source, and ideally why you open sourced this project, was to make something that was true to your vision and also incorporated other folk's ideas, needs and workflows. I understand that for a lot of folks, deleting 95% of a directory won't be a feature requirement. But for me, it is. I'm happy to do the work to implement it and hopefully we can come up with a PR that satisfies both your comfort and my workflow.

dragonpaw:
I ended up adding a confirmation popup, with the usual Yes/No, but also a checkbox you can select to "Don't ask me again this session" if you want to bulk delete like a madman, i.e. me.

luisangelsm:
I think I closed this by mistake, but I still need to think about it.

Issue #435, opened 2024-05-27
Feature Request: MacOS Touchpad gesture support


leu1:
I would like to request the addition of touchpad gesture support for YACReader on macOS. Implementing multi-touch gestures would greatly enhance the reading experience by making navigation and zooming in more intuitive and seamless. Thank you for considering this improvement.

PR #530, opened 2026-04-22
Add trackpad pinch-to-zoom gesture to the comic viewer

codeKonami:
Relates to #435.

## What this adds

Native trackpad pinch-to-zoom in the comic viewer:

- **Continuous zoom** from 30% to 500% (the existing bounds used by `Cmd +`/`Cmd -`) driven by the trackpad pinch gesture.
- **Cursor-anchored zoom.** The viewport is scrolled so the point under the cursor at the start of the gesture stays under the cursor while scaling, rather than zooming around the viewport center.
- **Compact bottom-right HUD** showing the live percentage while the gesture is active, so it does not obscure the page being read. The existing centered notification is preserved for keyboard `Cmd +` / `Cmd -` zoom.
- **Preferences toggle.** A new *Zoom mode* group is added to *Preferences > General*, below *Mouse mode*, with a single checkbox to enable or disable the gesture. Defaults to enabled.

## How it works

- `Viewer` now grabs `Qt::PinchGesture` in its constructor and overrides `event()` to dispatch `QEvent::Gesture` to a new `gestureEvent(QGestureEvent *)`.
- The gesture handler captures the zoom value and the cursor position in content coordinates at `Qt::GestureStarted`, then on `Qt::GestureUpdated` applies `baselineZoom * pinch->totalScaleFactor()` clamped to `[30, 500]`, resizes the content, and adjusts the horizontal/vertical scroll bars to keep the anchor stationary.
- The HUD is a dedicated `QLabel` child of the viewer, rendered as Qt rich text to keep the "%" text reliably white over the translucent background regardless of palette cascade. It shows on `GestureStarted` and hides on `GestureFinished` / `GestureCanceled`.
- The setting is persisted as `PINCH_TO_ZOOM_ENABLED` in the existing `QSettings` file, exposed via `Configuration::getPinchToZoomEnabled()` / `setPinchToZoomEnabled()`, and checked on each gesture event so toggling in Preferences takes effect immediately.

## Cross-platform

No `#ifdef Q_OS_MACOS` is needed. `QPinchGesture` is handled natively by Qt on macOS (trackpad) and Windows (touch), and is a no-op on platforms that do not deliver pinch events, so the feature is additive and does not regress existing behavior on Linux or on systems without a touch surface.

## Tested

- Built with Qt 6.11 on macOS 15 (Apple Silicon).
- Verified: gesture zooms and dezooms smoothly, cursor-anchored point stays under the cursor, `Cmd +` / `Cmd -` shortcuts and their HUD still work, two-finger scroll and the magnifying glass still work, the *Zoom mode* checkbox enables/disables the gesture at runtime.
- `clang-format` passes on the modified files.
- No change to existing public API, and no new runtime dependency.

Both Qt 5 and Qt 6 code paths should compile (`QPinchGesture` and `QGestureEvent` are available in both), though I have only been able to test Qt 6 locally.


luisangelsm:
Thank you, and sorry for taking me so long to review this.

I added some improvements and I removed the setting, I don't think a setting is needed for this.

PR #540, opened 2026-05-11
refactor(docker): Consolidate Dockerfile.aarch64 into Dockerfile

nickv2002:
## Summary

- After #505 lands, `docker/Dockerfile` and `docker/Dockerfile.aarch64` are byte-identical — same base (`linuxserver/baseimage-ubuntu:resolute`, a multi-arch manifest), same package lists, same multi-stage layout.
- This PR deletes `docker/Dockerfile.aarch64` and points the existing `docker-arm64` CI job at the shared `Dockerfile`.
- Pins `--platform linux/arm64` on the arm64 job to match the explicit `--platform linux/amd64` already used by the amd64 job. Both jobs still run on their native runners (`ubuntu-latest` and `ubuntu-24.04-arm`) and the manifest-combine step downstream is untouched, so multi-architecture publishing keeps working.

## Dependency

Stacked on **#505**. Must merge after #505 — the byte-identical state only exists once #505 lands. While #505 is open, this PR's diff will include the #505 commits; GitHub will narrow it down automatically after #505 merges.

## Test plan

- [x] Verified locally on Apple Silicon: native arm64 build from the consolidated `Dockerfile` produces a working image with `kimg_avif.so` + `kimg_jxl.so` present; `avif.cbz` and `jxl.cbz` pages render end-to-end through the HTTP API.
- [ ] After CI runs: confirm both `docker-amd64` and `docker-arm64` jobs succeed building from the single `Dockerfile`.
- [ ] Confirm the existing `docker buildx imagetools create` manifest-combine step still produces a working multi-arch tag.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


luisangelsm:
CI seems to pass.

luisangelsm:
@nickv2002 Thanks!

PR #544, opened 2026-05-19
fix: remove world writable permissions from install folder and binary

timbretimber:
Currently the installer creates the app directory and binaries with world-writable permissions. This allows any unprivileged user to delete or otherwise modify the files. The default install location is  under `%ProgramFiles%` which is generally considered privileged and should only be written to by Administrators. Even if installing to another directory, using inherited permissions should cover any reasonable use case.

Removing these additional permissions lets the directory and binaries inherit the the permissions set on `%ProgramFiles%`.


luisangelsm:
@timbretimber Thank you!!

PR #563, opened 2026-07-22
Update Chinese translations in yacreaderlibrary_zh_CN.ts

kofzhanganguo:
Update Simplified Chinese language pack


luisangelsm:
Thanks for the fixes!!

PR #566, opened 2026-07-24
Magnifying glass improvements (macOS)

starlit-void:
Two tiny bugfixes in one PR:

1. Magnifying glass zoom level appears to jump as the glass approaches the edge of the page. This is because of how the glass contents were painted, on hi resolution screens. Needs regression testing on non-macOS systems.
1. ~~Default +/- key events were stepping on the main window's +/- events when the magnifier was open. This prevented any of these key events from being consumed. Fixed by toggling off the +/- page events when magnifier is active. Note: if you set your page zoom to some other key combo, it may have worked while magnifier is open, but not after this fix. Not sure if this is an issue or not.~~
1. Ignore the page-selection bar at the bottom of the screen when the magnifying glass is active. This didn't show the page selection but it kept the magnifier from reaching the bottom of the page.

Passed clang-format, tested on macOS. I'm happy to make any changes if needed.
AI was used but I Know What I'm Doing™️.

Thank you!

EDIT 2026-07-25: Rewrote the branch to remove the shortcut toggles now that @luisangelsm fixed the issue upstream by changing the default keys. No need to disable the actions any more. I left the other two fixes as they were before.

### DIFF
--- a/YACReader/mouse_handler.cpp
+++ b/YACReader/mouse_handler.cpp
@@ -109,7 +109,9 @@ void YACReader::MouseHandler::mouseMoveEvent(QMouseEvent *event)
                     if (gtfPos.y() < 0 || gtfPos.x() < 0 || gtfPos.x() > viewer->goToFlow->width()) // TODO this extra check is for Mavericks (mouseMove over goToFlowGL seems to be broken)
                         viewer->animateHideGoToFlow();
                     // goToFlow->hide();
-                } else {
+                } else if (!viewer->magnifyingGlassShown) {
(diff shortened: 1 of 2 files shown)


luisangelsm:
I changed the default shortcuts here: https://github.com/YACReader/yacreader/commit/57b46fac7078ec6507f9c6f47b0659b82bf1881e

I am still not sure if we should disable the actions or not :)

starlit-void:
hi @luisangelsm, I removed the action toggles. thanks for the upstream fix!!

PR #567, opened 2026-07-24
Organize files

Andrey-Raspopov:
This adds a simple action to folder context menu that organizes files based on the metadata of the comics. The mask has several meaningful tokens, but the list can be expanded, I've tried to make it as safe as possible, and capable of removing empty folders on its own as well.


luisangelsm:
I gave it a try and I think it needs a full validation step showing exactly the output (ideally as a tree), some comics can end misplaced if they don't have the values for the fields used in the new path/name.

I can iterate over this...

Andrey-Raspopov:
I can work based on this feedback and caveats that I've noticed myself, I suppose that editable preview would make it thousand times better than just preview as well.

Andrey-Raspopov:
I've added editable preview, and padding for issue numbers, it works, but still can use more polish

luisangelsm:
Thank you, I'll think a bit more about the UX/UI.

luisangelsm:
BTW, I prefer rebasing to merging from develop.

Andrey-Raspopov:
Gotcha, will keep that in mind!

luisangelsm:
I disabled the feature behind a flag until it's ready for production

PR #575, opened 2026-07-31
Add configurable Escape key behavior (quit vs cancel mode)

starlit-void:
Escape previously always quit the viewer. Add an option, selectable via a radio group in Options ▸ General ▸ "Escape key", to instead make Escape cancel the topmost active mode and do nothing when none is active. Default stays "quit", so existing behavior is unchanged until opted into.

The Escape key and the File▸Close command are decoupled: closeAction keeps the configurable CLOSE_ACTION_Y/Esc binding but now dispatches through onEscapePressed() (quit vs. cancelActiveMode()), while a new exitAction backs the File▸Close menu item and always quits. This avoids making the menu command context-sensitive and avoids a second Esc binding (which would ambiguously collide).

cancelActiveMode() cancels in order: magnifier, dictionary/translator, go-to-flow bar, fullscreen. The magnifier's checkable toolbar action is re-synced when hidden this way. Viewer gains translatorIsVisible() and goToFlowIsVisible() accessors.

The cancel order is documented to users in a tooltip rather than in the radio button label, keeping the label as short as its neighbours in the General tab. The tooltip is a single translatable block with a translator comment: building the list from per-item tr() calls would leave word order, punctuation and numbering unfixable for translators, and markup inside a translatable string invites broken tags in the .ts files. Comments on both cancelActiveMode() and the tooltip flag that the two orders must stay in sync.

closeAction gets an explicit tooltip because the shortcuts editor lists actions by toolTip(), which otherwise falls back to the action text - showing "Escape" in a column whose neighbours are all verb phrases, next to a KEYS column that already names the key.

This is a quality-of-life usability improvement that serves me (starlit) so I understand if it isn't on the roadmap.


luisangelsm:
I rebased it on top of the new settings dialog + translations.

PR #577, opened 2026-07-31
macOS: don't require "control computer" accessibility permissions to move cursor

starlit-void:
When the user presses `S` to open the page flow, the cursor is warped to the page flow area. this is performed in Qt using `QCursor::setPos`, which internally synthesizes a mouse event, which requires "Allow the application to control your computer" accessibility setting on macOS. the proposed fix uses CGWarpMouseCursorPosition which moves the cursor without firing any events, which is basically what we want.

this is arguably a workaround for a Qt bug, not a YACReader bug.


starlit-void:
I've opened a Qt bug for this issue:
https://qt-project.atlassian.net/browse/QTBUG-148709

PR #581, opened 2026-08-02
Update translation for unsupported EPUB message

kofzhanganguo:
Update Simplified Chinese language translation


luisangelsm:
Thanks!!

BTW, you can create a single PR with both files updated/fixed.

PR #589, opened 2026-08-16
Fixed slow linux wayland mouse wheel scrolling in grid view

Luke-239:
Fixed extremely slow mouse wheel scrolling in grid view on Linux Wayland. Click and drag and scrollbar were unaffected. 

Confirmed the WheelHandler attached to GridView never received wheel events. It was still possible to scroll but whatever it fell back on did not work well, it required scrolling many more times (>10x) to move the same distance compared to the windows version.

Flow View already used MouseArea with an onWheel handler instead of WheelHandler and that worked so I changed grid view to use the same qml element for the main grid scroll and the info panel scroll. I was able to keep the actual scroll calculation exactly the same. I used the parent object size and filtered the other mouse buttons.

### DIFF
--- a/YACReaderLibrary/qml/GridComicsView.qml
+++ b/YACReaderLibrary/qml/GridComicsView.qml
@@ -533,14 +533,19 @@ SplitView {
                     currentIndexHelper.setGridColumnCount(wholeCells)
                 }
 
-                WheelHandler {
-                    onWheel: {
-                        if (grid.contentHeight <= grid.height) {
-                            return;
-                        }
+                MouseArea {
+                    anchors.fill: parent
+                    acceptedButtons: Qt.NoButton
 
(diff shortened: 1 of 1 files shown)


luisangelsm:
A really weird one. Thanks for investigating this. What version of Qt where you using?

luisangelsm:
@Luke-239 This breaks scrolling in subviews like the continue reading list and the current comic banners.

luisangelsm:
Fixed here: https://github.com/YACReader/yacreader/commit/02cab88df6216332675399fcf13a1a6c7f1f70f2

Luke-239:
@luisangelsm Sorry I should have included this info upfront. And I apologize for breaking the scrolling subviews. Thanks for merging it and fixing. 

KDE Plasma Version: 6.7.4
Qt Verison: 6.11.2
Kernel Version: 7.1.8-1-cachyos

